PR #6, opened 2020-04-07
Fixing a bug of displaying error when there are no annotations done by users

JingJZ160:
There was a bug that when there are no annotations have been done or there are no intersections among all users. This change display some info to explain why there are no Kappa stats.

### DIFF
--- a/ar/data.py
+++ b/ar/data.py
@@ -8,7 +8,6 @@ from collections import defaultdict, Counter
 import logging
 
 import pandas as pd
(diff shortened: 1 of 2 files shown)


edu-gp:
Nice. Is it easy to add a unit test?

JingJZ160:
> Nice. Is it easy to add a unit test?

Added a unit test for the edge cases.

edu-gp:
LGTM!

PR #9, opened 2020-04-08
Change the pattern model to return the non-overlapping matches that maximizes the score

JingJZ160:
The pattern matcher returns a list of all matches and it seems they are always ordered in ascending order by the length of match. To be less dependent on the API, I added a few lines of code to sort the result in ascending order. The result has the same format but only contains the largest and last matching result. For example, if the patterns are:
```
my
dog is healthy
healthy and happy
```
and the text is `my dog is healthy and happy`, then the return value is:
```
[{'tokens': ['my', 'dog', 'is', 'healthy', 'and', 'happy'],
   'matches': [(0, 1, 'my'), (3, 6, 'healthy and happy')], 'score': 4.0 / 6}]
```
The maximize method should cover all edge cases and it is cleaned up. If it doesn't make sense to you, I can share a super verbose version that has all the redundant if-else statements. That is the longest if-else statements I've written in years...

### DIFF
--- a/inference/pattern_model.py
+++ b/inference/pattern_model.py
@@ -14,15 +14,15 @@ def _compare_matches(item1, item2):
     :return: comparison result where -1 means item1 < item2, 1 means item1 >
     item2 and 0 means item1 == item2
     """
-    if (item1[2] - item1[1]) < (item2[2] - item2[1]):
+    if item1[1] < item2[1]:
         return -1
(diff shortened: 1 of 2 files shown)


edu-gp:
If the sentence is "my dog is healthy", we want both "my dog" and "is healthy" to match.

JingJZ160:
ping :)

PR #13, opened 2020-04-09
Rewrite the kappa calculation

JingJZ160:
Rewrite the kappa matrix calculation. The unit test is tricky. I was using a `set.intersection` in the code, which created an unstable output that failed the test intermittently...

For users who have no overlap on the annotation, the output is nan by default. Let me know if we want a prettier display.

image

### DIFF
--- a/ar/data.py
+++ b/ar/data.py
@@ -19,7 +19,8 @@ from db import _task_dir
 # I chose to write it all on disk for now - we can change it to a db later.
 
 
-def save_new_ar_for_user(task_id, user_id, annotation_requests, clean_existing=True):
+def save_new_ar_for_user(task_id, user_id, annotation_requests,
+                         clean_existing=True):
     '''
     Save a list of new annotation requests for a user (annotator).
     Args:
@@ -218,17 +219,20 @@ def compute_annotation_statistics(task_id):
         _get_all_annotators_from_annotated(task_id)
     )
 
-    total_distinct_annotations = _gather_labeled_examples(task_id)
-    annotations_per_user = dict()
+    total_distinct_annotations = _gather_distinct_labeled_examples(task_id)
+    anno_ids_per_user = dict()
+
+    results_per_task_user_anno_id = dict()
 
     for user_id in user_ids:
         anno_ids = fetch_all_annotations(task_id, user_id)
-        annotations_per_user[user_id] = set(anno_ids)
+        anno_ids_per_user[user_id] = set(anno_ids)
         n_annotations_per_user[user_id] = len(anno_ids)
 
         # TODO: slow
         for anno_id in anno_ids:
             anno = fetch_annotation(task_id, user_id, anno_id)
+            results_per_task_user_anno_id[(task_id, user_id, anno_id)] = anno
             for label, result in anno['anno']['labels'].items():
                 n_annotations_per_label[label][result] += 1
 
@@ -239,7 +243,8 @@ def compute_annotation_statistics(task_id):
 
     # kappa stats calculation
     kappa_table_per_label = _calculate_per_label_kappa_stats_table(
-        task_id, user_ids, annotations_per_user)
+        task_id, user_ids, anno_ids_per_user,
+        results_per_task_user_anno_id)
 
     return {
         'total_annotations': sum(n_annotations_per_user.values()),
@@ -247,13 +252,15 @@ def compute_annotation_statistics(task_id):
         'n_annotations_per_user': n_annotations_per_user,
         'n_annotations_per_label': n_annotations_per_label,
         'kappa_table_per_label': kappa_table_per_label,
-        'total_outstanding_requests': sum(n_outstanding_requests_per_user.values()),
+        'total_outstanding_requests': sum(
+            n_outstanding_requests_per_user.values()),
         'n_outstanding_requests_per_user': n_outstanding_requests_per_user,
     }
 
 
 def _calculate_per_label_kappa_stats_table(task_id, user_ids,
-                                           annotations_per_user):
+                                           anno_ids_per_user,
+                                           results_per_task_user_anno_id):
     """Calculate per label kappa matrix stats.
 
     Input structure of the annotations_per_user:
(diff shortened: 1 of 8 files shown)


ar/data.py
```diff
@@ -254,130 +253,177 @@ def compute_annotation_statistics(task_id):
...
-            anno = fetch_annotation(task_id, user_id, anno_id)
-            for label, result in anno['anno']['labels'].items():
-                kappa_stats_raw_data[label][user_id].append(result)
+    all_pairs_of_users = list(itertools.combinations(user_ids, 2))
+
+    for user1, user2 in all_pairs_of_users:
+        annotations_user1 = annotations_per_user[user1]
+        annotations_user2 = annotations_per_user[user2]
+
+        annotation_intersection = set.intersection(annotations_user1,
```
edu-gp, line 334: You need to remove any annotation that is "0" (unknown). ie. only keep 1 and -1.

ar/data.py
```diff
@@ -254,130 +253,177 @@ def compute_annotation_statistics(task_id):
...
+                                                   annotations_user2)
+        # WE NEED TO SORT THIS. OTHERWISE WE GOT UNSTABLE OUTPUT WHICH FAILS
+        # UNIT TESTS!
+        for anno_id in sorted(annotation_intersection):
+            anno_by_user1 = fetch_annotation(task_id, user1, anno_id)
+            anno_by_user2 = fetch_annotation(task_id, user2, anno_id)
```
edu-gp, line 340: This is potentially loading the same user's annotations one by one many times... even though we don't care that much about speed, ideally we just load it once. (Imagine if we were using a real DB, we'd want to do a batch SELECT. Let's code it like that.)

Below, I'll point you to where we're loading in all annotation data.

ar/data.py
```diff
@@ -479,7 +525,7 @@ def _export_labeled_examples(annotations_iterator):
     return final
 
 
-def export_labeled_examples(task_id, outfile=None):
+def _gather_labeled_examples(task_id):
     def annotations_iterator():
```
edu-gp, line 529: Following up to an earlier comment. Can we refactor and reuse this iterator to calculate Kappa? That is, load all the data with this function once.

It currently returns just the annotation, but you can change it to return (user_id, anno).

ar/data.py
```diff
@@ -254,130 +253,177 @@ def compute_annotation_statistics(task_id):
...
+    for label, df in kappa_matrices.items():
+        kappa_html_tables[label] = df.to_html(classes='kappa_table',
+                                              float_format=float_formatter)
+    return kappa_html_tables
+
+
+class PrettyDefaultDict(defaultdict):
+    """An wrapper around defaultdict so the print out looks like
+    a normal dict."""
+    __repr__ = dict.__repr__
```
edu-gp, line 424: Is this just for debugging?

ar/data.py
```diff
@@ -254,130 +253,177 @@ def compute_annotation_statistics(task_id):
...
     :param task_id: the id of a task
     :param user_ids: the user id list
-    :param annotations_from_all_users: annotations from all users
+    :param annotations_per_user: annotations per user dictionary
     :return: the per label kappa matrix html table
     """
     if len(user_ids) == 1:
         return ['There is only one user {}'.format(list(user_ids)[0])]
-    if len(annotations_from_all_users) == 0:
+    if len(annotations_per_user) == 0:
```
edu-gp, line 273: What if all user annotations are 0's (unknowns)?

ar/data.py
```diff
@@ -254,130 +253,177 @@ def compute_annotation_statistics(task_id):
...
         task_id,
         user_ids,
-        annotation_intersection
+        annotations_per_user
     )
-
-
-    logging.info(kappa_stats_raw_data)
-    kappa_matrices = _compute_kappa_matrix(user_ids, kappa_stats_raw_data)
+    kappa_matrices = _compute_kappa_matrix(kappa_stats_raw_data)
```
edu-gp, line 280: I'm curious. Do you think we can replace `_construct_per_label_per_user_pair_result` and `_compute_kappa_matrix` with a few nested for loops inside this function?

ar/data.py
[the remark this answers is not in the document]
JingJZ160: Will do. Didn't know we need to.

ar/data.py
JingJZ160: Will do.

ar/data.py
JingJZ160: Same as above.

ar/data.py
JingJZ160: Nope. It's an override so the logs will look cleaner on defaultdict-like object.

ar/data.py
JingJZ160: That is not covered as I didn't know the requirement to exclude 0s. I'll add checks for that.

ar/data.py
JingJZ160: The way GitHub lays out the code prevents me from knowing which function you are referring to. I'll keep that in mind. These two functions are extracted out since it is easier to do unit tests.

edu-gp: `_calculate_per_label_kappa_stats_table`, the function you're calling those 2 functions from.

ar/data.py
edu-gp: `annotations_iterator` inside `export_labeled_examples`, in `ar/data.py`

ar/data.py
JingJZ160: On a second thought, how about we just cache the call of `fetch_annotation`? My main concern of using the `_gather_labeled_examples` method is that it may take some time to refactor not just that method but potentially also all the calculation logic plus all the unit tests since they depends on a certain input data structure.

edu-gp: Sounds good

edu-gp: ^ as discussed on Slack, let's skip this for now.

ar/data.py
JingJZ160: I'm keeping these two functions separated. Otherwise it becomes a mess to write unit tests.

ar/data.py
JingJZ160: All unknowns will be excluded during the kappa calculation. If they are all zeros, then the score will be shown as `nan`.

JingJZ160:
PR updated. Kappa score calculation now includes labeling results only when both results are 1 or -1. If either is 0, both will be skipped. Fetch annotations are not invoked during the kappa calculation any more. I added a dictionary with key of (task_id, user_id, annotation_id) so we can do a look up. It's not a huge improvement but at least it serves as a in memory lookup table without the LRU cache issue we discussed.

JingJZ160:
Just did some profiling on the B2C example so we can refer to it in the future (ms on average):
- `fetch_annotation` takes about 6 ms for each `(task, user, anno_id)`
- `fetch_all_annotation_ids` takes about 8 to 10 ms for each `(task, user)`
- `fetch_all_ar` takes about 80 ms.
- `_export_distinct_labeled_examples` is invoked once but takes about 3000 ms alone.

ar/data.py
```diff
@@ -405,7 +492,7 @@ def _export_labeled_examples(annotations_iterator):
                 }
             },
             'anno': {
-                'labels': {
+                'labels': {ƒ
```
edu-gp, line 495: Please remove this special character.

JingJZ160: Done.

PR #15, opened 2020-04-16
Integration with SqlAlchemy plus the temporary REST API for Label and EntityType

JingJZ160:
- Integration with SqlAlchemy using `flask-alchemy` and migration tool using `alembic`. 
- Test DB framework using `flask-testing`
- REST API for Label and EntityType

Unit tests are incomplete. I'll add more in a different PR.


alembic/versions/740c83f3631f_first_commit.py
```diff
@@ -0,0 +1,54 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('entity_type',
+    sa.Column('id', sa.Integer(), nullable=False),
+    sa.Column('name', sa.String(length=64), nullable=True),
+    sa.PrimaryKeyConstraint('id')
```
edu-gp, line 24: nit: Can we add another indentation? makes it easier to read. Ditto for the other function calls here as well.

alembic/versions/740c83f3631f_first_commit.py
```diff
@@ -0,0 +1,54 @@
...
+    )
+    op.create_index(op.f('ix_entity_type_name'), 'entity_type', ['name'], unique=True)
+    op.create_table('entity',
+    sa.Column('id', sa.Integer(), nullable=False),
+    sa.Column('name', sa.String(length=64), nullable=True),
+    sa.Column('entity_type_id', sa.Integer(), nullable=True),
+    sa.ForeignKeyConstraint(['entity_type_id'], ['entity_type.id'], ),
+    sa.PrimaryKeyConstraint('id')
+    )
+    op.create_index(op.f('ix_entity_name'), 'entity', ['name'], unique=True)
```
edu-gp, line 34: If we're already putting a unique index on the name, do we still need an id? (Maybe name can just function as the primary key?)

It seems it would be cleaner to code, but it does mean all foreign keys pointing to an Entity needs to store a string instead of an id... not sure what the best practice here is.

ar/data.py
```diff
@@ -19,6 +20,33 @@
 # I chose to write it all on disk for now - we can change it to a db later.
 
 
+def fetch_labels_by_entity_type(entity_type_name):
+    entity_type = EntityType.query.filter_by(name=entity_type_name).first()
+    if entity_type:
+        labels = [label.name for label in entity_type.labels.all()]
```
edu-gp, line 26: Is this making 2 calls to the db?
```
EntityType.query.filter_by(name=entity_type_name).first()
```
and 
```
entity_type.labels.all()
```

Is there a way to phrase it with a join so it just make a single call?

ar/data.py
```diff
@@ -19,6 +20,33 @@
...
+        return labels
+    else:
+        return []
+
+
+def save_labels_by_entity_type(entity_type_name, labels):
+    logging.error("Finding the EntityType for {}".format(entity_type_name))
+    entity_type_id = db.session.query(EntityType.id).filter_by(
+        name=entity_type_name).scalar()
+    if entity_type_id is None:
```
edu-gp, line 36: You don't need to do this check. You're already using a unique index on the entity's name.

ar/data.py
```diff
@@ -19,6 +20,33 @@
...
+        db.session.add(entity_type)
+        db.session.commit()
+        entity_type_id = entity_type.id
+        logging.error("Created a new EntityType {} with id {}".format(
+            entity_type_name, entity_type_id
+        ))
+    for label_name in labels:
+        label = Label(name=label_name, entity_type_id=entity_type_id)
+        db.session.add(label)
+    db.session.commit()
```
edu-gp, line 47: There are 2 calls to `db.session.commit()` Do you need both?

db/model.py
```diff
@@ -0,0 +1,32 @@
...
+    id = db.Column(db.Integer, primary_key=True)
+    name = db.Column(db.String(64), index=True, unique=True)
+    entity_type_id = db.Column(db.Integer, db.ForeignKey('entity_type.id'))
+    fake_field = db.Column(db.String(64), nullable=True)
+
+
+class EntityType(db.Model):
+    id = db.Column(db.Integer, primary_key=True)
+    name = db.Column(db.String(64), index=True, unique=True)
+    # TODO how should we setup the lazy loading mode?
```
edu-gp, line 18: Is the TODO still relevant? Looks like you set up lazy-loading here already.

docker/Dockerfile
```diff
@@ -26,5 +26,7 @@ RUN wget -nv \
     # Remove the backup directory that gcloud creates
     rm -rf /root/tools/google-cloud-sdk/.install/.backup
 
+RUN pip install flask-sqlalchemy alembic Flask-Testing
```
edu-gp, line 29: Missing SQLAlchemy here, and please move this up to where all the other pip installs are.

docker/run.sh
```diff
@@ -1,4 +1,7 @@
 echo "Starting Server"
 
 cd /annotation_tool
-supervisord -n -c docker/supervisord.conf
\ No newline at end of file
+# We probably need to add some code to run database migration here before
+# starting the actual server.
```
edu-gp, line 5: It's okay. Let's do migrations manually for now!

docker/supervisord.conf
```diff
@@ -49,12 +49,14 @@ autorestart=true
...
-environment=FLASK_APP='backend',FLASK_ENV="development"
+environment=FLASK_APP='backend',FLASK_ENV="development",
+DATABASE_URL="annotation_tool/alchemy.db"
 priority=1000
 
 [program:frontend]
 command=flask run -h 0.0.0.0 --port 5001
-environment=FLASK_APP='frontend',FLASK_ENV="development"
+environment=FLASK_APP='frontend',FLASK_ENV="development",
+DATABASE_URL="annotation_tool/alchemy.db"
```
edu-gp, line 59: We shouldn't include the repo name here. If it's just "alchemy.db" instead of "annotation_tool/alchemy.db", does that work?

frontend/__init__.py
```diff
@@ -13,24 +16,31 @@
 def create_app(test_config=None):
     # create and configure the app
     app = Flask(__name__, instance_relative_config=True)
+    # TODO how do we add credentials for sqlite and specify them in the config
```
edu-gp, line 19: I don't think we need credentials for now. But if we do, let's pass them in through environment variables and read them out with `os.environ`

frontend/__init__.py
```diff
@@ -13,24 +16,31 @@
...
     app.config.from_mapping(
         SECRET_KEY='athena_todo_change_this_in_prod',
         DATABASE=os.path.join(app.instance_path, 'athena.sqlite'),
     )
 
+    logging.error(DevelopmentConfig.SQLALCHEMY_DATABASE_URI)
```
edu-gp, line 25: Why logging.error?

frontend/config.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class DevelopmentConfig(Config):
+    DEBUG = True
+
+
+class TestingConfig(Config):
+    DEBUG = True
+    TESTING = True
+    SQLALCHEMY_DATABASE_URI = 'sqlite:///:memory:'
```
edu-gp, line 25: Can we put this in a shared directory? Maybe the `db` directory. This database isn't just for the frontend :)

frontend/labels.py
```diff
@@ -0,0 +1,40 @@
...
+    labels = set(fetch_labels_by_entity_type(entity_type_name))
+    logging.error("Existing labels: {}".format(labels))
+    new_labels = new_labels.difference(labels)
+    try:
+        logging.error("Updating new labels {} for EntityType {}".format(
+            str(new_labels), entity_type_name
+        ))
+        save_labels_by_entity_type(entity_type_name, list(new_labels))
+        msg = "Labels for entity {} have been updated".format(entity_type_name)
+        return Response(msg, status=200, mimetype='application/json')
```
edu-gp, line 37: Oh interesting. I've always used this instead: http://flask.pocoo.org/docs/latest/api/#flask.json.jsonify Just FYI!

alembic/versions/740c83f3631f_first_commit.py
JingJZ160: It's auto-generated but I'll add more indentation.

ar/data.py
JingJZ160: I can give a try.

ar/data.py
JingJZ160: We need an entity_type_id to create labels. I don't know if that id is available without the first commit. I can give it a try.

db/model.py
JingJZ160: Lazy loading has a lot of modes actually. I just picked one for the test. It is not necessary the best option.

docker/Dockerfile
JingJZ160: SQLAlchemy is part of flask-sqlalchemy.

docker/supervisord.conf
JingJZ160: I don't know, probably need to change a few places to test out.

frontend/__init__.py
JingJZ160: Just a line of test code. I'll remove it. I use error mostly because I want to see info log and error log in one place.

frontend/__init__.py
```diff
@@ -13,24 +16,31 @@
...
+        app.config.from_object(test_config)
 
     # ensure the instance folder exists
     try:
         os.makedirs(app.instance_path)
     except OSError:
         pass
 
+    db.init_app(app)
+    # migrate.init_app(app=app)
```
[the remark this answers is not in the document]
JingJZ160: Not any more. I'll remove.

frontend/__init__.py
```diff
@@ -54,6 +64,20 @@ def secret():
...
+    # conn = sqlite3.connect('alchemy.db')
+    # c = conn.cursor()
+    #
+    # # Create table
+    # # c.execute('''DROP TABLE IF EXISTS labels''')
+    # c.execute('''CREATE TABLE IF NOT EXISTS labels
+    # (entity varchar(30), labels json)''')
+    #
+    # conn.commit()
+    # conn.close()
```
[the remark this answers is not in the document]
JingJZ160: Yep. Will remove.

frontend/config.py
JingJZ160: For sure. I was only developing for the frontend at the beginning.

frontend/labels.py
JingJZ160: Can we specify status code in jsonify?

frontend/start_server.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+export FLASK_APP=frontend
+export FLASK_ENV=development
+#flask db init
+#flask db migrate
+#flask db upgrade
+flask run -h 0.0.0.0 --port 5000
```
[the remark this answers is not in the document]
JingJZ160: Oops. I thought I removed it.

tests/unit/test_frontend_model.py
```diff
@@ -0,0 +1,37 @@
...
+        db.drop_all()
+
+    def populate_db(self):
+        entity_type = EntityType(name="company")
+        db.session.add(entity_type)
+        db.session.commit()
+        label1 = Label(name="B2C", entity_type_id=entity_type.id)
+        label2 = Label(name="HEALTHCARE", entity_type_id=entity_type.id)
+        db.session.bulk_save_objects([label1, label2])
+        db.session.commit()
```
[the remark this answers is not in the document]
JingJZ160: Same as above. Need to test.

ar/data.py
JingJZ160: Why not? If the entity_type doesn't exist, we need to create it first.

ar/data.py
edu-gp: Yea makes sense. I guess you do need 2 commits.

alembic/versions/740c83f3631f_first_commit.py
JingJZ160: Well it's auto-generated. We don't have to write it ourselves.

frontend/labels.py
edu-gp: I believe you can do this:
```
response = jsonify(error.to_dict())
response.status_code = ...
return response
```
Yuck, so many lines. 

Btw, I just realized what you're returning here is a string, not a json. (nit!)

It's fine. Feel free to leave as is - we'll have to formally decide what this thing should return later.

ar/data.py
JingJZ160: Yep, failed unit test without the first commit. We need the commit to get the id.

tests/unit/test_frontend_model.py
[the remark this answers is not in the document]
JingJZ160: Failed test without the first commit. We need it to get the entity type id.

ar/data.py
edu-gp: Yea actually on second thought I think what you're doing is right. (I was thinking just insert every time and it'll fail when the entity already exists - but I believe it's faster to read, and plus you need the entity_type_id below)

frontend/__init__.py
JingJZ160: Then I'll add a TODO here as a reminder.

alembic/versions/740c83f3631f_first_commit.py
edu-gp: On second thought I think this is okay. Storing the entity type as a string in every entity & label seems silly

But if we were to do it, it looks like if you just do:
```
name = db.Column(db.String(64), primary_key=True)
```
Maybe the auto-generated migration would take care of it...

JingJZ160: Yeah I'd rather keep the EntityType table, in case we need to change the name of any entity type later.

ar/data.py
JingJZ160: It's doable.

docker/supervisord.conf
JingJZ160: Hmm somehow it is not working without the repo name. I'll look into it tomorrow. It's probably related to the working directory issue.

JingJZ160: Somehow a new alchemy.db is created under the frontend folder, which is weird...

docker/Dockerfile
edu-gp: Please move this up to where all the other pip installs are.

db/model.py
```diff
@@ -0,0 +1,31 @@
...
+
+db = SQLAlchemy()
+metadata = db.Model.metadata
+
+
+class Label(db.Model):
+    id = db.Column(db.Integer, primary_key=True)
+    name = db.Column(db.String(64), index=True, unique=True)
+    entity_type_id = db.Column(db.Integer, db.ForeignKey('entity_type.id'))
+    fake_field = db.Column(db.String(64), nullable=True)
```
edu-gp, line 11: Can you create another migration to remove this fake field?

frontend/__init__.py
```diff
@@ -13,24 +16,31 @@
 def create_app(test_config=None):
     # create and configure the app
     app = Flask(__name__, instance_relative_config=True)
+    # TODO add credentials for sqlite, probably from os.environ
     app.config.from_mapping(
         SECRET_KEY='athena_todo_change_this_in_prod',
-        DATABASE=os.path.join(app.instance_path, 'athena.sqlite'),
+        # DATABASE=os.path.join(app.instance_path, 'athena.sqlite'),
```
edu-gp, line 22: Delete this line please - no longer need it :)

JingJZ160: Done.

docker/Dockerfile
JingJZ160: Done.

PR #17, opened 2020-04-17
Adding context, user, annotation tables and add one-to-many relationships

JingJZ160:
Adding context, user and annotation tables. Please check the one-to-many relationship between tables.

Another thing we should think about is on the delete cascade. For example, one EntityType can have many Labels associated but do we want to delete all labels when the EntityType is deleted? By default it's not. The foreign key on the label object pointing to the entity type is just set to Null.

What about user and annotation? When we delete a user, should we delete all the annotations the user has touched?


db/model.py
```diff
@@ -24,7 +27,65 @@ def __repr__(self):
...
+    annotations = db.relationship('Annotation', backref='user', lazy='dynamic')
+
+    def __repr__(self):
+        return '<User {}>'.format(self.username)
+
+
+class Context(db.Model):
+    id = db.Column(db.Integer, primary_key=True)
+    hash = db.Column(db.String(128), index=True, unique=True)
+    data = db.Column(db.TEXT)
```
edu-gp, line 52: Can we use the JSON type?
https://docs.sqlalchemy.org/en/13/core/type_basics.html#sqlalchemy.types.JSON

JingJZ160: Sure thing. I was hesitant on which one to choose so I followed the `text` type in the design but I'm OK with JSON.

PR #19, opened 2020-04-18
Add Task and BackgroundJob

edu-gp:
This is being developed on top of https://github.com/edu-gp/annotation_tool/pull/18 <- we need to merge that first.

This adds the Task and BackgroundJob objects. A BackgroundJob can be to generate annotation requests, or to train a model, or to take a snapshot of the data, etc.

This actually covers pretty much (the majority of, if not all of) the rest of the database we need for now. After this we can start migrating over the data and logic.


alembic/versions/4b41206a16ea_add_task_and_backgroundjob.py
```diff
@@ -0,0 +1,54 @@
...
+    )
+    op.create_table('background_job',
+    sa.Column('id', sa.Integer(), nullable=False),
+    sa.Column('type', sa.Integer(), nullable=False),
+    sa.Column('params', sa.JSON(), nullable=False),
+    sa.Column('output', sa.JSON(), nullable=False),
+    sa.Column('status', sa.String(length=64), nullable=False),
+    sa.Column('created_at', sa.DateTime(timezone=True), server_default=sa.text('(CURRENT_TIMESTAMP)'), nullable=True),
+    sa.Column('updated_at', sa.DateTime(timezone=True), nullable=True),
+    sa.Column('task_id', sa.Integer(), nullable=True),
```
JingJZ160, line 37: Please reformat the code.

edu-gp:
@JingJZ160 A background job is like a python script, it can run independently of the flask server.

JingJZ160:
> @JingJZ160 A background job is like a python script, it can run independently of the flask server.

Okay if that's the type of background job we are running, then this extension does not apply.

edu-gp:
Will close this and merge it into master.

PR #22, opened 2020-04-20
Migration script from loading the file to the db tables

JingJZ160:
Script for loading the file to the db tables. Please run the script under the project root instead of the bin folder like `python bin/migrate_json_to_db.py` but make sure you edit the script to specify the task ids. Thought this could be more flexible than fully automate loading the whole `__task` folder.

There is something missing in the python library path but it seems that we need to do some bigger renaming and folder moving changes to fix it. Since this is a one-time thing (I hope), I'm just hard-coding the path. 

----Update----
Run with `python -m scripts.migrate_json_to_db` under the project root folder.

### DIFF
--- a/ar/data.py
+++ b/ar/data.py
@@ -1,3 +1,4 @@
+import copy
 import os
 import shutil
 import itertools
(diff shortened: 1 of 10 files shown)


ar/data.py
```diff
@@ -191,11 +191,12 @@ def fetch_annotation(task_id, user_id, ar_id):
         return None
 
 
-def fetch_all_annotation_ids(task_id, user_id):
+def fetch_all_ar_ids(task_id, user_id):
     '''
     Return a list of ar_id for this task that has been annotated by this user.
     '''
     _dir = os.path.join(_task_dir(task_id), DIR_ANNO, user_id)
+    print(_dir)
```
edu-gp, line 199: remove

shared/config.py
```diff
@@ -8,30 +8,35 @@ class Config:
 
     This module is designed to fail loudly when a required env var is not found.
     """
-
-    def get_backend_password():
+    @classmethod
```
edu-gp, line 11: `@staticmethod` ? Since `cls` is not used.

bin/migrate_json_to_db.py
```diff
@@ -0,0 +1,85 @@
+import copy
```
edu-gp, line 1: We already have a `scripts` folder, can we move this file under there?

bin/migrate_json_to_db.py
```diff
@@ -0,0 +1,85 @@
...
+
+sys.path.append(".")
+
+from ar import fetch_all_ar_ids
+from ar.data import fetch_annotation, _get_all_annotators_from_annotated
+from db.config import DevelopmentConfig
+from db.model import Database, User, EntityType, Context, Entity, Label, \
+    ClassificationAnnotation
+
+ENTITY_TYPE_NAME = "company"
```
edu-gp, line 15: Can we keep the entity type in an enum like https://github.com/edu-gp/annotation_tool/blob/master/db/model.py#L146

bin/migrate_json_to_db.py
```diff
@@ -0,0 +1,85 @@
...
+
+def _convert_single_annotation(anno, username):
+    user = get_or_create(User, username=username)
+
+    entity_type = get_or_create(EntityType, name=ENTITY_TYPE_NAME)
+
+    annotation_context = json.dumps(anno["req"]["data"], sort_keys=True)
+    context = get_or_create(Context, exclude_keys_in_read=["data"],
+                            hash=hashlib.md5(
+                                annotation_context.encode()).hexdigest(),
```
edu-gp, line 35: Let's refactor this hash function for reuse

ar/data.py
JingJZ160: Done.

shared/config.py
JingJZ160: Done.

bin/migrate_json_to_db.py
JingJZ160: Done.

bin/migrate_json_to_db.py
JingJZ160: Sure. We've already have `EntityType` so I'll name it as `EntityTypeEnum`.

bin/migrate_json_to_db.py
JingJZ160: Done.

PR #23, opened 2020-04-21
Add AnnotationRequest

edu-gp:
1. Add AnnotationRequest model
2. Cleanup style in db/model.py
3. Add unit tests. I had to name the unit tests with `zzz` at the front to make sure they run at the end. If they're run before a test that uses `flask_testing`, the `flask_testing` tests seem to reuse the same database I use. Not sure why this is happening, will dig deeper later.


db/model.py
```diff
@@ -87,6 +120,34 @@ class User(Base):
     def __repr__(self):
         return '<User {}>'.format(self.username)
 
+    def fetch_ar_count_per_task(self):
+        """Returns a list of tuples (name, task_id, count)
+        """
+        session = inspect(self).session
```
JingJZ160, line 126: Why not just `db.session`? Why do we need `inspect` here?

db/model.py
```diff
@@ -87,6 +120,34 @@ class User(Base):
...
+            .filter(AnnotationRequest.user == self) \
+            .group_by(Task.id) \
+            .order_by(Task.id)
+
+        return q.all()
+
+    def fetch_ar_for_task(self, task_id,
+                          status=AnnotationRequestStatus.Pending):
+        """Returns a list of AnnotationRequest objects"""
+        session = inspect(self).session
```
JingJZ160, line 141: Same as above.

tests/unit/test_zzz_annotation_request_model.py
```diff
@@ -0,0 +1,86 @@
+from db.model import (
+    Database, AnnotationRequest, BackgroundJob, JobType, JobStatus,
+    User, Context, Task, AnnotationType
+)
+from db.config import TestingConfig
+
+
+class TestAnnotationRequestModel:
```
JingJZ160, line 8: What is the error if this test is not executed in the last place? I still find it weird that the order matters here since the `teardown` method should clean up everything even if the same database is connected...

db/model.py
edu-gp: `db` is a Flask-Sqlalchemy thing. We use inspect to get the _current_ database session the instance is using.

tests/unit/test_zzz_annotation_request_model.py
edu-gp: The culprit is `Base.query` --> https://github.com/edu-gp/annotation_tool/blob/master/db/model.py#L28

That's the thing that enables you to do `User.query.all()` instead of `session.query(User).all()` in Flask-Sqlalchemy.

The teardown does work, but it looks like that `.query` is still pointing to the previous connection.

I believe this is only an issue in tests.

db/model.py
JingJZ160: OK if we plan to use this method outside the flask app in the background job, then `db` is not the right option. Feel free to merge the branch.

PR #25, opened 2020-04-21
Refactor code for Label and EntityType

edu-gp:
1. Moved code about Label and EntityType from `ar/data` to `db/model`, because those classes are not specific to annotation requests.
2. I added a `dbsession` param to `fetch_labels_by_entity_type` and `save_labels_by_entity_type` so the user is forced to pass in a session. Doing so removes dependecy on Flask-Sqlalchemy at the expense of being more explicit.

I made the changes in the frontend tests. @JingJZ160 if you like, I can also remove the dependency on `flask_testing` for that test.

### DIFF
--- a/db/model.py
+++ b/db/model.py
@@ -1,3 +1,5 @@
+import logging
+import copy
 from flask_sqlalchemy import SQLAlchemy
 from sqlalchemy import create_engine, inspect
 from sqlalchemy.schema import ForeignKey, Column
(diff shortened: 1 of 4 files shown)


JingJZ160:
>     1. Moved code about Label and EntityType from `ar/data` to `db/model`, because those classes are not specific to annotation requests.
> 
>     2. I added a `dbsession` param to `fetch_labels_by_entity_type` and `save_labels_by_entity_type` so the user is forced to pass in a session. Doing so removes dependecy on Flask-Sqlalchemy at the expense of being more explicit.
> 
> 
> I made the changes in the frontend tests. @JingJZ160 if you like, I can also remove the dependency on `flask_testing` for that test.

Yes could you remove the dependency for that test so we stay consistent? Thanks.

db/model.py
```diff
@@ -281,3 +283,61 @@ class AnnotationRequest(Base):
 
     # Where this request came from. e.g. {'source': BackgroundJob, 'id': 123}
     source = Column(JSON)
+
+
+# =============================================================================
+# Convenience Functions
+
+
+def get_or_create(db_session, model, exclude_keys_in_retrieve=None, **kwargs):
```
JingJZ160, line 292: Maybe rename `db_session` to `dbsession` (or the opposite) for consistency?

PR #27, opened 2020-04-21
Stats calculation

JingJZ160:
Use db query to calculate stats for annotations per label. Well, some calculation are easier to do in code. 

Stats for request will be in another PR. 

This is also a breaking changing for the UI. Wiring with UI will be in another PR too.

### DIFF
--- a/ar/data.py
+++ b/ar/data.py
@@ -12,7 +12,7 @@ from typing import List
 import pandas as pd
 from pandas import DataFrame
 from sklearn.metrics import cohen_kappa_score
-from sqlalchemy import func
+from sqlalchemy import func, distinct
 
 from db import _task_dir
 from db.model import db, EntityType, Label, User, ClassificationAnnotation, \
@@ -247,10 +247,6 @@ def compute_annotation_statistics(task_id):
         n_outstanding_requests_per_user[user_id] = len(
             set(ar_ids) - set(anno_ids))
 
-    # kappa stats calculation
-    # kappa_table_per_label = _calculate_per_label_kappa_stats_table(
-    #     task_id, user_ids, anno_ids_per_user,
-    #     results_per_task_user_anno_id)
     kappa_table_per_label = None
 
     return {
@@ -278,20 +274,20 @@ def compute_annotation_statistics_db(dbsession, label_name):
         for num, user in num_of_annotations_done_per_user
     }
 
-    annotations = dbsession.query(ClassificationAnnotation).all()
+    total_distinct_annotations = dbsession.query(
+        ClassificationAnnotation).count()
 
     # kappa stats calculation
-    distinct_users = set([
-        annotation.user for annotation in annotations
-    ])
-    kappa_stats_raw_data = _construct_kappa_stats_raw_data(distinct_users,
-                                                           label_name)
+    distinct_users = dbsession.query(distinct(
+        ClassificationAnnotation.user)).all()
+    kappa_stats_raw_data = _construct_kappa_stats_raw_data(
+        db.session, distinct_users, label_name)
 
     kappa_table_per_label = _compute_kappa_matrix(kappa_stats_raw_data)
 
     return {
         'total_annotations': total_num_of_annotations_done_by_users,
-        'total_distinct_annotations': len(annotations),
+        'total_distinct_annotations': total_distinct_annotations,
         'n_annotations_per_user': n_annotations_done_per_user_dict,
         'kappa_table_per_label': kappa_table_per_label,
     }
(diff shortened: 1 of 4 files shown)


ar/data.py
```diff
@@ -258,107 +265,96 @@ def compute_annotation_statistics(task_id):
...
-                "user_id2": [-1, 1]
-            },
-            ...
-        },
-        ...
+
+def _compute_total_annotations(dbsession, label_name):
+    num_of_annotations_done_per_user = dbsession.query(
+        func.count(ClassificationAnnotation.id),
+        User.username). \
```
edu-gp, line 303: Can we move the bracket below and indent these 2 lines? Make it a little easier to read.

```
num_of_annotations_done_per_user = dbsession.query(
        func.count(ClassificationAnnotation.id),
        User.username
    ). \
```

Personally I also like to put the `.` on the next line, but I'm ok either way.

ar/data.py
```diff
@@ -242,9 +248,10 @@ def compute_annotation_statistics(task_id):
             set(ar_ids) - set(anno_ids))
 
     # kappa stats calculation
-    kappa_table_per_label = _calculate_per_label_kappa_stats_table(
-        task_id, user_ids, anno_ids_per_user,
-        results_per_task_user_anno_id)
+    # kappa_table_per_label = _calculate_per_label_kappa_stats_table(
+    #     task_id, user_ids, anno_ids_per_user,
+    #     results_per_task_user_anno_id)
```
edu-gp, line 253: Can these comments be removed?

ar/data.py
```diff
@@ -258,107 +265,96 @@ def compute_annotation_statistics(task_id):
...
-    kappa_stats_raw_data = _construct_per_label_per_user_pair_result(
-        task_id,
-        user_ids,
-        anno_ids_per_user,
-        results_per_task_user_anno_id
-    )
-    kappa_matrices = _compute_kappa_matrix(kappa_stats_raw_data)
-    kappa_matrix_html_tables = _convert_html_tables(kappa_matrices)
-    return kappa_matrix_html_tables
+    annotations = dbsession.query(ClassificationAnnotation).all()
```
edu-gp, line 281: Do you need to load in all the fields of annotation?

ar/data.py
```diff
@@ -258,107 +265,96 @@ def compute_annotation_statistics(task_id):
...
 
+    # kappa stats calculation
+    distinct_users = set([
+        annotation.user for annotation in annotations
```
edu-gp, line 285: Is this firing a query for every single annotation? Would it suffice to use `user_id` instead?

If not, maybe you can change the query above (`annotations = dbsession.query(ClassificationAnnotation).all()`) to load in all the users.

ar/data.py
```diff
@@ -258,107 +265,96 @@ def compute_annotation_statistics(task_id):
...
+    return kappa_stats_raw_data
 
-    all_pairs_of_users = list(itertools.combinations(user_ids, 2))
 
-    for user1, user2 in all_pairs_of_users:
-        annotations_user1 = annos_per_user[user1]
-        annotations_user2 = annos_per_user[user2]
+def _retrieve_annotation_with_same_context_shared_by_two_users(user1, user2):
+    annotations_from_user1 = user1.classification_annotations
+    annotations_from_user2 = user2.classification_annotations
```
edu-gp, line 327: Oh my god. So this is doing 2 queries for each pair of users... let's say sqlalchemy is smart enough to cache the calls, this is still sending N_USERS number of queries. Am I understanding this right?

Is it possible to load in everything needed to compute kappa with 1 call?

(Ideally we do all these calculations in sql, but that can wait until later...)

ar/data.py
JingJZ160: Done.

ar/data.py
JingJZ160: Done.

ar/data.py
JingJZ160: User instances are needed. I'll change it to query the users directly.

ar/data.py
JingJZ160: I guess not.

ar/data.py
JingJZ160: Good point. Similar to our original code, I've refactored the logic to load all the context_ids and annotation_values for the users we care about and saved them in a dictionary. This provides a lookup table to retrieve the annotations in this method without querying the database repeatedly. 

I have no idea how to write the whole kappa calculation in one complex SQL query that is readable (and easy to understand in the future)......That's why I wrote this in code in the first place.

ar/data.py
```diff
@@ -258,107 +261,138 @@ def compute_annotation_statistics(task_id):
...
+    total_distinct_annotations = dbsession.query(
+        ClassificationAnnotation).count()
 
-def _construct_per_label_per_user_pair_result(task_id, user_ids,
-                                              annos_per_user,
-                                              results_per_task_user_anno_id):
-    """Construct the per label per user_pair labeling result dictionary.
+    # kappa stats calculation
+    distinct_users = dbsession.query(distinct(
+        ClassificationAnnotation.user)).all()
```
edu-gp, line 282: All users, or just all users who has annotated with this label?

Can you pull users from the `num_of_annotations_done_per_user` object above instead of making another call?

JingJZ160: Done, although I have to add another small utility class.

PR #28, opened 2020-04-22
Add DataSnapshotJob

edu-gp:
Using [Single Table Inheritance](https://docs.sqlalchemy.org/en/13/orm/inheritance.html#single-table-inheritance) I created `DataSnapshotJob`, `AnnotationRequestJob`,  `TextClassificationModelTrainingJob`, and `TextClassificationModelInferenceJob` as children of `BackgroundJob`.

This is nice because we don't have to manage job types manually and still enjoy the benefits of having a single table & shared behaviour.

I implemented the behaviour of `DataSnapshotJob` here, the other tables will follow in future PRs.

-- Edit --
I needed to fix the Alembic migration, change the `BackgroundJob:type` field from an Integer to a String. Please delete the local db and run migration again.

### DIFF
--- a/db/model.py
+++ b/db/model.py
@@ -195,7 +195,7 @@ class BackgroundJob(Base):
     __tablename__ = 'background_job'
 
     id = Column(Integer, primary_key=True)
-    type = Column(String, index=True, nullable=False)
+    type = Column(String(64), index=True, nullable=False)
     params = Column(JSON, nullable=False, default=lambda: {})
     output = Column(JSON, nullable=False, default=lambda: {})
     status = Column(String(64), nullable=False, default=JobStatus.Init)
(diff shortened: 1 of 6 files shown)


db/model.py
```diff
@@ -208,10 +207,85 @@ class BackgroundJob(Base):
...
+
+    @staticmethod
+    def create(label_name: str, now=None):
+        """Convert all the annotations associated with this label into data
+        ready to be used externally.
+        """
+        if now is None:
+            now = int(time.time())
+
+        output_fname = f"{now}_{secure_filename(label_name)}.jsonl"
```
JingJZ160, line 232: What does secure_filename do?

edu-gp: It converts it to a safe string for to be stored on disk. Eg. "A/bad_label" -> "A_bad_label". See more : https://werkzeug.palletsprojects.com/en/1.0.x/utils/#werkzeug.utils.secure_filename

PR #30, opened 2020-04-23
Changes made to wire up with the UI of the backend server

JingJZ160:
Code changes required to wire things up for the backend server UI. Please see the screenshot to check if this makes sense. 

The data shown here are from the task `8a79a035-56fa-415c-8202-9297652dfe75` loaded in the database. Since all the requests have the status of Pending, the outstanding requests are shown as 400 (and 100 per user). 

The rest of the page (Annotations and Model) is shown per label.  We only have one label B2C so only one of these are displayed. 

The basic info section are still extracted from a file.  Are they still valid? If not, what should we change?

image

### DIFF
--- a/ar/data.py
+++ b/ar/data.py
@@ -16,8 +16,8 @@ from db import _task_dir
 from db.model import db, Label, User, ClassificationAnnotation, \
     AnnotationRequest, AnnotationRequestStatus
 from db.task import Task, DIR_ANNO, DIR_AREQ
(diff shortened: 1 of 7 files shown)


JingJZ160:
> LGTM! The only comment is, in the "Annotations for B2C" box, can you move the stuff on the right (Request More Annotations, etc.) and the stuff below ("Annotator Login Links") to the "Annotation Requests" box? Those belong with requests, not any specific label.

Done. 
image

PR #33, opened 2020-04-24
Associate annotation with annotation_request

JingJZ160:
A few things touched:
- We need to add automatic constraint naming to the metadata, otherwise the migration script may fail when adding foreign keys. https://alembic.sqlalchemy.org/en/latest/naming.html
- Annotation request now has a reference to an annotation_id. Since not all annotation_requests are annotated in our existing data, only part of the requests will have non-null annotation_id.
- Migration script has been updated to create the annotation request and its related annotation together.


db/model.py
```diff
@@ -12,7 +12,14 @@
...
 from db.fs import MODELS_DIR, TRAINING_DATA_DIR
 
-Base = declarative_base()
+meta = MetaData(naming_convention={
+        "ix": "ix_%(column_0_label)s",
+        "uq": "uq_%(table_name)s_%(column_0_name)s",
+        "ck": "ck_%(table_name)s_%(constraint_name)s",
+        "fk": "fk_%(table_name)s_%(column_0_name)s_%(referred_table_name)s",
+        "pk": "pk_%(table_name)s"
+      })
```
edu-gp, line 21: Why do we need to specify our own naming convention?

Can we remove this if we just wipe all the alembic migrations and generate a new one? (Since we still have the luxury to do this!)

JingJZ160: We could wipe out the db and redo everything but what if we need to modify something again? Somehow the without the naming convention, the migration script failed on the creation of a foreign key. In fact, it would be great if you can try it on your end for verification. Just add the foreign key like I did and run the migration script. Does it fail?

edu-gp: Yep it fails (And if it's a nullable=False it also fails with this meta thing... for other reasons). I guess it doesn't hurt to add it.

JingJZ160: OK. Let's keep it for now. It's strange that the doc says that it's built-in...Well, at least the convention is copied from the doc directly...

PR #34, opened 2020-04-24
Remove ModelViewer

edu-gp:
- Delete ModelViewer, move its logic into either db.model::Model or db.model::FileInference.
- Wire up the UI so we can still view the`index` and `show` pages for Tasks.

Jing this sets you up well to rewire the `create` and `edit` pages for Tasks.


backend/tasks.py
```diff
@@ -213,27 +212,31 @@ def train(id):
...
+
+        # Write it to a temp file and send it.
+        import tempfile
+        import os
+        from werkzeug.utils import secure_filename
         from flask import send_file
-        return send_file(final_fname, mimetype='text/csv', cache_timeout=0, as_attachment=True)
+
+        with tempfile.TemporaryDirectory() as tmpdirname:
+            name = f"{secure_filename(task.name)}__{stem(fname)}.csv"
```
JingJZ160, line 233: Since we are accessing `.name` of a `task`, should we add a check on whether `task` is none?

db/task.py
```diff
@@ -211,21 +211,24 @@ def parse_jinjafied(field, value):
...
-        if mv is not None:
-            return NLPModel(self.task_id, mv.version)
-        else:
-            return None
+        raise Exception("Deprecated Function")
+        # mv = self.get_active_model_viewer()
+        # if mv is not None:
+        #     return NLPModel(self.task_id, mv.version)
+        # else:
+        #     return None
```
JingJZ160, line 234: Do we still need those comments?

shared/utils.py
```diff
@@ -89,3 +89,20 @@ class PrettyDefaultDict(defaultdict):
...
 def gen_uuid():
     return str(uuid.uuid4())
+
+
+def file_len(fname):
+    try:
+        with open(fname) as f:
+            for i, l in enumerate(f):
+                pass
+        return i + 1
```
JingJZ160, line 99: Wait...variable `i` is still visible outside the `for` loop in Python?

edu-gp: Yep

PR #37, opened 2020-04-27
Add ClassificationTrainingData.create_for_label to export annotations

edu-gp:
This PR allows us to export annotations from the database, and put it in a file ready to be consumed by the model trainer.

We do this through a new method `ClassificationTrainingData.create_for_label`, which takes a Label and gathers all the annotations for that label with a SQL query. `create_for_label` also requires a function to return text for an entity - this is something to be injected at runtime, and its implementation will be in a later PR.

Load tests show we can handle a lot annotations very quickly - any reasonable size file takes less than a second.


edu-gp:
Edit: Need to redo this after any potential (likely) schema changes in https://github.com/edu-gp/annotation_tool/pull/38

PR #38, opened 2020-04-28
Experimental changes to the db to better support labeling

edu-gp:
Per discussion on https://github.com/edu-gp/annotation_tool/pull/36 ,  here are my proposed schema changes. **See the test for a story**, check if it covers all the cases.

Details:

- I added `entity` and `label` foreign keys to `AnnotationRequest`.
- I changed the `context` in `AnnotationRequest` to a JSON, and removed the `Context` object entirely. The idea is once an Annotation is created, we can **copy** the entire context json object to the Annotation.

If this is what we want, there are a few more things I need to add to this PR:

- [x] Remove code involving Context
- [x] Drop Context table, update Alembic
- [x] Change migration script
- [x] Fix all the tests
- [x] Update how I'm gathering annotations for training --> will do in https://github.com/edu-gp/annotation_tool/pull/37


tests/unit/test_annotation_flow.py
```diff
@@ -0,0 +1,79 @@
...
+
+def test_annotation_flow(dbsession):
+    entity_type = EntityType(name='Fruit')
+    user = User(username='Bob')
+    dbsession.add_all([user, entity_type])
+    dbsession.commit()
+
+    # When we set up an annotation request, we create the following, as needed:
+    label = Label(name="IsSweet", entity_type=entity_type)
+    entity = Entity(name="Banana", entity_type=entity_type)
```
JingJZ160, line 17: Where and when should we create them? During `Request more annotations`?

tests/unit/test_annotation_flow.py
```diff
@@ -0,0 +1,79 @@
...
+
+    # [RECEIVE_ANNOTATION] <- user_id, req_id, entity_name, label_name, value
+    # Bob suddenly remembers something else about Bananas.
+    label_is_yellow = Label(name="IsYellow", entity_type=entity_type)
+    anno = ClassificationAnnotation(
+        label=label_is_yellow, value=1,
+        user=user, entity=entity, context=req.context)
+    # Note we copied the context over from the AnnotationRequest.
+    dbsession.add(anno)
+    dbsession.commit()
```
JingJZ160, line 70: I assume this works in a way that Bob is still on the `annotate` page and there is going to be another section to add additional labels that are not associated with the original request. He can add `yellow` as an additional label and must press `Next` to move on. Is this the same as you mentioned in the other PR?

db/model.py
```diff
@@ -163,19 +168,19 @@ def fetch_ar_for_task(self, task_id,
...
+#     data = Column(JSON, nullable=False)
+#     # A context can be part of many annotations.
+#     classification_annotations = relationship('ClassificationAnnotation',
+#                                               backref='context',
+#                                               lazy='dynamic')
 
-    def __repr__(self):
-        return '<Context {}: {}>'.format(self.id, self.data)
+#     def __repr__(self):
+#         return '<Context {}: {}>'.format(self.id, self.data)
```
JingJZ160, line 183: Without the context table, this requires us to send the whole context to `receive_annotation` POST API. Could you remind me why we decide to get rid of this table?

db/model.py
```diff
@@ -456,26 +461,24 @@ class AnnotationRequest(Base):
...
-    # What should the user annotate.
-    context_id = Column(Integer, ForeignKey('context.id'), nullable=False)
-    context = relationship("Context")
+    # What Entity should the user annotate.
+    entity_id = Column(Integer, ForeignKey('entity.id'), nullable=False)
+    entity = relationship("Entity")
+
+    # What Label we want the user to annotate.
+    label_id = Column(Integer, ForeignKey('label.id'), nullable=False)
+    label = relationship("Label")
```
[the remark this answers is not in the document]
edu-gp: The way I'm thinking about `AnnotationRequest` is that it's asking a single direct question. For example, "Jing, can you tell me if Bill Gates founded Microsoft?" If you answer that question, then you've fulfilled the `AnnotationRequest`. You're also free to tell me about things about Bill Gates, and that would be other Annotations (ie. multi-labeling)

Does that make sense?

tests/unit/test_annotation_flow.py
edu-gp: Yes

tests/unit/test_annotation_flow.py
edu-gp: Yes

db/model.py
edu-gp: Yea that's fine.

Imagine if the annotations now come from Slack, or a Chrome plugin, or a bulk upload from Salesforce, they all have to POST their contexts along with the annotation.

We could have the Context table, it just seems unnecessary to me. When would we need to query by Context?

JingJZ160: That's a fair point. If we want to receive annotations from outside the current system, we need to post the whole thing.

db/model.py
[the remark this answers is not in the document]
JingJZ160: Yes makes sense.

PR #39, opened 2020-04-28
Switch Task Edit & Create pages to use DB

edu-gp:
- Use DB when creating or editing a Task
- Added getters and setters to the Task model to get some common attributes (Jing you might find these useful)

### DIFF
--- a/tests/unit/test_task_model.py
+++ b/tests/unit/test_task_model.py
@@ -11,3 +11,29 @@ def _populate_db(dbsession):
 def test_sanity(dbsession):
     _populate_db(dbsession)
     assert len(dbsession.query(Task).all()) == 1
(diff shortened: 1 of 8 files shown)


tests/unit/test_task_model.py
```diff
@@ -3,11 +3,38 @@
...
 
 
 def test_sanity(dbsession):
     _populate_db(dbsession)
     assert len(dbsession.query(Task).all()) == 1
+
+
+def test_get_set(dbsession):
+    # Note: Saving any modifications to JSON requires
+    # marking them as modified with `flag_modified`.
```
JingJZ160, line 18: Maybe move this comment to the `Task` class?

PR #40, opened 2020-04-28
Switch NLPModel to use the database

edu-gp:
I switched the interface of the `NLPModel` constructor to `NLPModel(dbsession, task_id)`

See the unit test at the end for an example use of `NLPModel`.

### DIFF
--- a/tests/unit/test_nlp_model.py
+++ b/tests/unit/test_nlp_model.py
@@ -1,0 +1,70 @@
+from tests.sqlalchemy_conftest import *
+from db.model import (
+    Task, TextClassificationModel, FileInference
+)
+from inference.nlp_model import NLPModel
+from shared.utils import stem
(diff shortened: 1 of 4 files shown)


tests/unit/test_nlp_model.py
```diff
@@ -0,0 +1,70 @@
...
+
+    # =========================================================================
+    # Create mock files
+
+    model_uuid = "abc"
+    version = 1
+    data_fname = "myfile.csv"
+
+    # Save the data file
+    d = tmp_path / "raw_data"
```
JingJZ160, line 23: Wait is this another python black magic to create a file path?

edu-gp: I think this is something special to the https://docs.python.org/3/library/pathlib.html module

PR #42, opened 2020-04-29
Add a function to generate entity text lookup function

edu-gp:
Add `get_entity_text_lookup_function`, which is a function that returns a function that takes an (entity_type_id, entity_name) and returns the text associated with it. We use this to generate training data.

This function will be used in `ClassificationTrainingData.create_for_label`, which was introduced in https://github.com/edu-gp/annotation_tool/pull/37

### DIFF
--- a/shared/utils.py
+++ b/shared/utils.py
@@ -123,3 +123,17 @@ def textarea_to_list(text: str):
     res = [x.strip() for x in text.split('\n')]
     res = [x for x in res if len(x) > 0]
     return res
+
+
+def json_lookup(json_data, key):
+    """
+    Give a key "a.b.c", look up json_data['a']['b']['c']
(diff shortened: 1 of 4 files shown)


shared/utils.py
```diff
@@ -123,3 +123,17 @@ def textarea_to_list(text: str):
...
     res = [x for x in res if len(x) > 0]
     return res
+
+
+def json_lookup(json_data, key):
+    """
+    Give a key "a.b.c", look up json_data['a']['b']['c']
+    Returns None if any of the keys were not found.
+    """
+    sofar = json_data
```
JingJZ160, line 133: `sofar`? Why this particular variable name?

edu-gp: Because it's where you are in the dictionary so far.

PR #45, opened 2020-04-29
Request more annotations

JingJZ160:
Code changes for wiring up request generation. It is not fully tested. I want to get your feedback on the logic and see if they all make sense first, specifically the blacklist function and the cleanup of existing requests logic.

Most new methods have a suffix of `_db` since I don't want to break existing code yet.

### DIFF
--- a/ar/__init__.py
+++ b/ar/__init__.py
@@ -27,6 +27,7 @@ Pred_DB = namedtuple('Pred_DB', ['score', 'entity_name', 'fname',
 
 db = Database.from_config(DevelopmentConfig)
 
+
 def generate_annotation_requests(dbsession, task_id: int,
                                  max_per_annotator: int, max_per_dp: int):
     '''
@@ -88,11 +89,29 @@ def generate_annotation_requests(dbsession, task_id: int,
     # TODO basic logic, given a user_id, task_id, and entity_id, we should
     #  be able to find if there are exisiting requests for this user and
     #  entity under this task. If so, skip those.
-    blacklist_fn_db = _build_blacklist_fn_db(task=task)
+
+    num_of_complete_requests_by_entity_and_user_within_task = \
+        dbsession.query(
+            func.count(AnnotationRequest.id),
+            Entity.name,
+            User.username
+        ). \
+        join(Entity). \
+        join(User). \
+        filter(AnnotationRequest.task_id == task.id,
+               AnnotationRequest.status ==
+               AnnotationRequestStatus.Complete). \
+        group_by(Entity.name, User.username).all()
+
+    lookup_dict = {
+        (item[1], item[2]): item[0]
+        for item in num_of_complete_requests_by_entity_and_user_within_task
+    }
+
+    blacklist_fn_db = _build_blacklist_fn_db(lookup_dict=lookup_dict)
 
     logging.info("Assigning to annotators...")
-    assignments = _assign_db(dbsession,
-                             ordered_examples,
+    assignments = _assign_db(ordered_examples,
                              task.get_annotators(),
                              blacklist_fn=blacklist_fn_db,
                              max_per_annotator=max_per_annotator,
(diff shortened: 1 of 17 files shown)


ar/__init__.py
```diff
@@ -16,38 +23,50 @@
 Pred = namedtuple('Pred', ['score', 'fname', 'line_number'])
 
 
-def generate_annotation_requests(task_id: str,
+def generate_annotation_requests(dbsession, task_id: str,
```
edu-gp, line 26: task_id is an int now, not a str

ar/__init__.py
```diff
@@ -16,38 +23,50 @@
...
 
     print("Get prediction from each model...")
     # TODO better to stream these?
 
     _examples = []
     _proportions = []
 
+    data_filenames = [os.path.join(filestore_base_dir(), RAW_DATA_DIR,
+                                   fname) for fname in
+                      task.get_data_filenames()]
```
edu-gp, line 42: I'm going to push a PR soon, with that you can just call do `data_filenames  = task.get_data_filenames(abs=True)`

ar/__init__.py
```diff
@@ -16,38 +23,50 @@
...
+    # TODO The Pred returned from get_predictions are based on filename and
+    #  line numbers so I have to save a dictionary of entity name here for
+    #  the blacklist function, unless we have a better way.
+    #  One optimization we can do is to creat this while populate the cache
+    #  below since both have to read in all the files.
+    entity_per_file_per_line_dict = _build_entity_name_per_file_per_line(
```
edu-gp, line 48: I would much rather we change `get_predictions` to return predictions per entity name. It just makes more sense than using filename & line numbers.

ar/__init__.py
```diff
@@ -127,17 +157,53 @@ def get_decorated_example(pred: Pred):
...
+        if lookup_key not in _lookup_db:
+            res = dbsession.query(func.count(AnnotationRequest.id)). \
+                join(Entity). \
+                join(User). \
+                filter(Entity.name == entity_name,
+                       User.username == annotator,
+                       AnnotationRequest.task_id == task.id,
+                       AnnotationRequest.status ==
+                       AnnotationRequestStatus.Complete).one()[0]
+            _lookup_db[lookup_key] = res > 0
```
edu-gp, line 200: For a typical model (<10k annotations), I would guess it's faster to load all the (entity_name, username) pairs into memory than to query each of them one at a time. What do you think?

ar/__init__.py
```diff
@@ -176,6 +243,74 @@ def _get_predictions(data_filenames: List[str], models: List[ITextCatModel], cac
     return result
 
 
+def _assign_db(dbsession, entity_lookup: dict,
```
edu-gp, line 246: Does this function need `dbsession`? It's easier to test without it.

ar/data.py
```diff
@@ -31,6 +31,45 @@
...
+            # TODO not the most efficient way since it involves a query to get
+            #  the entity id.
+            new_request = AnnotationRequest(
+                user_id=user_id,
+                entity_id=dbsession.query(Entity.id).filter(Entity.name == req[
+                    'entity_name']).one()[0],
+                annotation_type=AnnotationType.ClassificationAnnotation,
+                task_id=task_id,
+                context=req['data'],
+                # TODO what about order? Self generated on insert?
```
edu-gp, line 63: `order` is the order to present things to the user. Looks like you can just use the loop index.

ar/data.py
```diff
@@ -31,6 +31,45 @@
...
+            )
+            dbsession.add(new_request)
```
edu-gp, line 65: You might want to try this with a few hundred requests per user, for up to 100 users, just to see 1) how fast is it and 2) if there are errors.

frontend/__init__.py
```diff
@@ -9,7 +9,7 @@
 from db.config import DevelopmentConfig
 from .auth import login_required
 
-from db.task import Task
+from db._task import _Task
```
edu-gp, line 12: I was just looking at this file. I believe you can delete this line

ar/__init__.py
JingJZ160: Done.

ar/__init__.py
JingJZ160: Probably not. They are not used in the function.

ar/__init__.py
JingJZ160: Technically the result is cached but yeah it will still be faster to do a bulk group_by and count query at the beginning.

ar/__init__.py
JingJZ160: Added another field for `entity_meta_data` which includes the `name` and `domain`.

JingJZ160:
OK the code works, except the requests don't have `label_id` and the `entity` doesn't have `entity_type_id` because the raw data don't contain label and entity type info. 

As for the performance, for the 2000 sample in the raw data and all the annotators we have, it was about 28 seconds. I'll add more users to do a stress test.

ar/data.py
JingJZ160: It takes less than 2 minutes for 100 new users with the spring_small.jsonl.

JingJZ160:
An overview of the performance:
For 200 requests (4 annotators and 50 requests per person):
- If we have new entities, it takes about 10 seconds in total, where 6 seconds are spent on running prediction from the PatternModel. The rest takes 4 seconds.
image

- If we rerun the process, it takes about 8 seconds, where 6 seconds are spent on predictions from the PatternModel. The rest takes 2 seconds.
image

So the bottleneck at this moment? The prediction from the PatternModel. The cache parameter is hardcoded to `False` by a TODO item so it's always executed.

I don't know what speed we should expect on the insert of the requests. We are using sqlite. Shall we try a different db? Where can I get the whole raw file from pitchbook? We can add a separate task for performance test.

PR #46, opened 2020-04-29
Model training with DB (Local)

edu-gp:
This PR adds the ability to train a model (local training only - I will follow up with a PR for remote training)

The main change is in the function `train/prep.py::prepare_task_for_training`. This function creates a model directory which contains a `config.json` and a `data.jsonl`, and performs any necessary database updates. The model directory is then passed to a local training function, or be copied over somewhere else for remote training.

This function currently hard-codes some things that are very specific to our current file formats. This will make everything work for now. Will fix later.

Other changes:
- The `Model` table now explicitly stores the training `config` and `classification_training_data`, instead of storing both in a json field. This is a little cleaner. I updated the migration script + included a Alembic migration.
- Fixed tests and deprecated old function.

Tested with `pytest` and tested by training a model locally.


edu-gp:
Merging now to unblock me

PR #47, opened 2020-04-30
Control database URI with an env var

edu-gp:


edu-gp:
> LGTM. It still works, right? I remember there was an issue with this that an extra db was created.

It doesn't change the previous logic. I usually set the env var to an abs path anyway so I don't think it should be a problem.

PR #49, opened 2020-04-30
Remote training with DB

edu-gp:
Changes to make training work on Google AI Platform. Mostly a lot of renaming.

Depends on https://github.com/edu-gp/annotation_tool/pull/48 first


edu-gp:
Will open and merge against db_logic_migration

PR #51, opened 2020-04-30
Fix bug where tests are using live DB

edu-gp:
https://app.clickup.com/t/6jcm8f


edu-gp:
Merging to unblock me

PR #54, opened 2020-05-01
Bug fixes

JingJZ160:
- Blacklisting function. Tested by annotating a website `whitesourcesoftware.com` and then request more annotations. Fresh frontend and go through the whole list. No sign of this website. Please double check on your end as well.
- Pattern info is still not shown on my frontend somehow and it complains about the info is null so I have to add additional check for that. Otherwise the text won't show up. It is a separate issue why the info is null in the first place.
- The next_req in the annotate method could be None if there is no next request and that causes an error on the last request page. Besides, we want to get the label for the current request, not the next...

### DIFF
--- a/ar/__init__.py
+++ b/ar/__init__.py
@@ -9,7 +9,8 @@ from sqlalchemy import func
 
 from db.fs import filestore_base_dir, RAW_DATA_DIR
 from db.model import get_or_create, Task, \
-    AnnotationRequest, User, AnnotationRequestStatus
+    AnnotationRequest, User, AnnotationRequestStatus, ClassificationAnnotation, \
+    AnnotationValue
 from shared.utils import load_jsonl
 from inference.base import ITextCatModel
 from inference import get_predicted
@@ -20,7 +21,7 @@ from .utils import get_ar_id, timeit
 
 # TODO eddie
 Pred = namedtuple('Pred', ['score', 'entity', 'fname',  'line_number'])
-UserEntityTaskTuple = namedtuple('UserEntityTaskTuple', ['user', 'entity'])
+EntityUserTuple = namedtuple('EntityUserTuple', ['entity', 'user'])
 
 
 def generate_annotation_requests(dbsession, task_id: int,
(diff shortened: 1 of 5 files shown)


ar/__init__.py
```diff
@@ -150,6 +137,24 @@ def generate_annotation_requests(dbsession, task_id: int,
...
+        ). \
+        join(User). \
+        filter(ClassificationAnnotation.value !=
+               AnnotationValue.NOT_ANNOTATED). \
+        group_by(ClassificationAnnotation.entity, User.username).all()
+    lookup_dict = {
+        EntityUserTuple(item[1], item[2]): item[0]
+        for item in num_of_annotations_by_entity_and_user_within_task
+    }
+    return lookup_dict
```
edu-gp, line 155: I'm totally okay with this, though in theory we don't need the count, unless I'm misunderstanding something? (Optional)

ar/data.py
```diff
@@ -228,11 +228,15 @@ def fetch_ar(task_id, user_id, ar_id):
...
+    res = dbsession.query(AnnotationRequest.id).filter(
         AnnotationRequest.task_id == task_id,
         AnnotationRequest.user_id == user_id,
         AnnotationRequest.id > current_ar_id
-    ).order_by(AnnotationRequest.id.asc()).first()[0]
+    ).order_by(AnnotationRequest.id.asc()).first()
+    if res is not None:
+        return res[0]
+    else:
+        return res
```
edu-gp, line 239: Just `return None` please.

frontend/tasks.py
```diff
@@ -60,7 +60,8 @@ def annotate(task_id, ar_id):
     # TODO get_next_ar_id_from_db should just return the AnnotationRequest
     next_req = db.session.query(AnnotationRequest) \
         .filter_by(id=next_ar_id).one_or_none()
-    label = next_req.label
+    # next_req could be None.
+    label = ar_dict['label']
```
edu-gp, line 64: Null next req is currently an undefined behaviour.... Maybe just redirect back to the index if there's nothing more to label?

JingJZ160: Eh we still want to annotate the current request, no? When a user click on a link, say `TODO - EXAMPLE 300`, it triggers `/annotate/300` and the website should show the annotate page for request 300 and try to fetch the next request. If request 300 is the last request (thus no next request), we can't just redirect back to the index page here...

PR #55, opened 2020-05-01
Remove FileInference

edu-gp:
Drop the FileInference table, it's really not needed - we don't query on it, and all its functionalities can be handled by the parent Model.


edu-gp:
Merging to unblock

PR #59, opened 2020-05-05
Bug fixes from prod

edu-gp:
Multiple bug fixes from prod + added some tests

1. Add `psycopg2-binary` as a dependency to connect to Postgres
2. Fix queries with groupby's to use all keys selected
3. Fix frontend to show AnnotationRequest in order
4. Make sure a task's active nlp model is one that has finished training

### DIFF
--- a/tests/unit/test_model.py
+++ b/tests/unit/test_model.py
@@ -2,6 +2,7 @@ from mockito import when
 from sqlalchemy.exc import ArgumentError
 
 from tests.sqlalchemy_conftest import *
+from tests.utils import fake_train_model
 from db.model import TextClassificationModel, get_or_create, User
 import pytest
 import sqlalchemy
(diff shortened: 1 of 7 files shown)


tests/unit/test_model.py
```diff
@@ -69,3 +70,18 @@ def test_get_or_create(dbsession):
...
                           username="invalid_user_name")
+
+
+def test_is_ready(dbsession, monkeypatch, tmp_path):
+    monkeypatch.setenv('ALCHEMY_FILESTORE_DIR', str(tmp_path))
+
+    model = TextClassificationModel(uuid='123', version=1)
+    dbsession.add(model)
+    dbsession.commit()
+    dbsession.refresh(model)
```
JingJZ160, line 81: Why do we need to refresh the model?

edu-gp: This just means you're reloading it from the database, just to make sure we're testing the actual data in the db.

PR #60, opened 2020-05-07
Fix bugs in scripts/migrate_json_to_db.py

edu-gp:
Fix a few bugs in prod that was causing data to be dropped when we migrated over.

### DIFF
--- a/scripts/migrate_json_to_db.py
+++ b/scripts/migrate_json_to_db.py
@@ -41,16 +41,21 @@ def convert_annotation_result_in_batch(task_uuid, username):
         print("Converted annotation with ar_id {}".format(ar_id))
 
 
-def _make_company_entity(meta_dict):
-    company_name = meta_dict.get("name", "unknown")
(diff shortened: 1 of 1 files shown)


scripts/migrate_json_to_db.py
```diff
@@ -170,16 +176,20 @@ def _convert_single_request_with_annotated_result(
...
+        # closed (e.g. restart all processes in supervisor)
         from db.model import Base
         Base.metadata.drop_all(db.engine)
         Base.metadata.create_all(db.engine)
 
     mock_time = 1000000000
 
-    # task_uuids = ["8a79a035-56fa-415c-8202-9297652dfe75"]
+    #task_uuids = ["fe9a1e62-80e3-4e58-9e12-3247ac0d18f5"]
+    #task_uuids = ["f29dca1f-f03b-4667-b7fd-643e6f1a4611"]
```
JingJZ160, line 188: Do we still need these two lines?

PR #70, opened 2020-05-12
Show a page to display all the data files

edu-gp:
This makes it easy for a user to verify a file was exported succesfully.

### DIFF
--- a/backend/data.py
+++ b/backend/data.py
@@ -1,0 +1,31 @@
+from flask import (
+    Blueprint, flash, redirect, render_template, request, url_for
+)
+
+from db.utils import get_all_data_files, get_all_pattern_files
+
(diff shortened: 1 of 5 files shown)


backend/data.py
```diff
@@ -0,0 +1,31 @@
...
+from db.utils import get_all_data_files, get_all_pattern_files
+
+from .auth import auth
+
+bp = Blueprint('data', __name__, url_prefix='/data')
+
+
+@auth.login_required
+def _before_request():
+    print("Before Request")
```
JingJZ160, line 14: Do we need to print this message or are we planning to show something else here?

edu-gp: lol good catch

PR #71, opened 2020-05-12
First version of the script

JingJZ160:
This is actually a work in progress. I have some TODO items in the script that could use some feedback before we implement the code. 

I'm only inserting new annotations if there are no annotations created by Salesforce bot for a website with targeted labels. This is based on the assumption that we won't unmark a website from targeted to non-targeted in Salesforce.

For the other query that look at LeadStatus, do we also need a time range constraint? The Lead table only has company name but no company domain. The account table may have domain but not company name. If we use a join, it may not return the right data. BTW, I couldn't get the join working in Salesforce SQL so it is two separate queries plus code. 

Also if we want to use this script, we probably need a secret manager on GCP as it's not safe to check in the permissions here.

Let me know your thoughts.


scripts/pull_salesforce.py
```diff
@@ -0,0 +1,207 @@
...
+        AND CreatedDate < {}
+        """.format(selected_companies_sql_condition,
+                   start_time_str,
+                   end_time_str)
+    )
+    return result['records']
+
+
+# TODO are we ever going to unmark an entity from B2C to non-B2C by
+#  salesforce? If no, then there is no point of update existing entries.
```
edu-gp, line 131: Ideally, if there's a duplicate in Salesforce, for whatever reason, let's just take the latest value as the truth.

But the way you're doing it is fine too. If there's duplicate in labels, the downstream training script will just take a majority vote.

scripts/pull_salesforce.py
```diff
@@ -0,0 +1,207 @@
...
+def insert_salesforce_b2c_annotations(
+        dbsession,
+        salesforce_user: User,
+        last_dropoff_reasons: List,
+        filter_by: str = "Business Model - B2C") -> NoReturn:
+    entities_dropped_due_to_b2c = [
+        record["cbit__ClearbitDomain__c"]
```
edu-gp, line 138: Do clearbit domains always exist? Do you ever want to fall back to the website (and extract domain yourself)?

scripts/pull_salesforce.py
```diff
@@ -0,0 +1,207 @@
...
+        model=User,
+        username="salesforce_bot"
+    )
+    insert_salesforce_b2c_annotations(
+        dbsession=db.session,
+        salesforce_user=salesforce_bot,
+        last_dropoff_reasons=records
+    )
+
+
```
edu-gp, line 205: Add a `db.session.close()` at the end

scripts/pull_salesforce.py
```diff
@@ -0,0 +1,207 @@
...
+# TODO are we ever going to unmark an entity from B2C to non-B2C by
+#  salesforce? If no, then there is no point of update existing entries.
+def insert_salesforce_b2c_annotations(
+        dbsession,
+        salesforce_user: User,
+        last_dropoff_reasons: List,
+        filter_by: str = "Business Model - B2C") -> NoReturn:
+    entities_dropped_due_to_b2c = [
```
edu-gp, line 137: You're already making `filter_by` a parameter, might as well make this variable name more general as well.

JingJZ160: Done.

scripts/pull_salesforce.py
JingJZ160: Done.

scripts/pull_salesforce.py
```diff
@@ -0,0 +1,207 @@
...
+def fetch_lead_status(api: Salesforce,
+                      start_time: datetime,
+                      end_time_exclusive: datetime) -> List:
+    start_time_str = start_time.strftime('%Y-%m-%dT%H:%M:%SZ')
+    end_time_str = end_time_exclusive.strftime('%Y-%m-%dT%H:%M:%SZ')
+
+    companies_with_bd_accepted = api.query_all(
+        """
+        SELECT Lead.Company FROM Lead WHERE 
+        Lead.Status = 'BD Accepted' Limit 10
```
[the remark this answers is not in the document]
JingJZ160: There are only 5 values for this field:
- Pending Review
- BD Accepted
- Deferred
- Dead
- Already Exists

So I think using just "BD Accepted" is enough here. The rest do not tell us if they are B2C or not.

scripts/pull_salesforce.py
JingJZ160: Now we are using upsert.

scripts/pull_salesforce.py
JingJZ160: Added a function to do extraction but I'm skipping companies who do not have domain or website. The name of such companies are logged in a warning messages for now. Let me know if this is something we want.

ar/data.py
```diff
@@ -237,8 +237,7 @@ def get_next_ar_id_from_db(dbsession, task_id, user_id, current_ar_id):
     ).order_by(AnnotationRequest.id.asc()).first()
     if res is not None:
         return res[0]
-    else:
-        return res
+    return None
```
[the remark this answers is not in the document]
JingJZ160: Sure.

docker/Dockerfile
```diff
@@ -28,4 +28,10 @@ RUN wget -nv \
...
 # Path configuration
-ENV PATH $PATH:/root/tools/google-cloud-sdk/bin
\ No newline at end of file
+ENV PATH $PATH:/root/tools/google-cloud-sdk/bin
+
+#RUN apt-get install cron
+#
+#COPY crontab.txt /var/crontab.txt
+#RUN crontab /var/crontab.txt
+#RUN chmod 744 /etc/crontab
```
edu-gp, line 37: Can we not commit these until we actually decide to use cron?

docker/Dockerfile
```diff
@@ -4,8 +4,8 @@ RUN apt-get update
 RUN apt-get install -y redis-server vim
 
 RUN pip install redis pytest Flask supervisor Celery seaborn simpletransformers python-dotenv Flask-HTTPAuth
-RUN pip install mockito
-RUN pip install flask-sqlalchemy alembic Flask-Testing psycopg2-binary
+RUN pip install mockito simple-salesforce PyJWT cryptography requests
+RUN pip install flask-sqlalchemy alembic Flask-Testing psycopg2-binary tldextract
```
edu-gp, line 8: Can we put the new packages on a new line, and indicate that these are the new requirements for Salesforce integration?

scripts/pull_salesforce.py
```diff
@@ -0,0 +1,209 @@
...
+        'beta': '.beta',
+        'prod': ''
+    }
+    # ENV = 'prod'
+    IS_SANDBOX = env != 'prod'
+
+    consumer_key = "To be added"
+    KEY_FILE = 'To be added'
+    ISSUER = consumer_key
+    SUBJECT = 'To be added' + ENV_SUFFIX_MAP[env]
```
edu-gp, line 41: How are you going to set these?

docker/Dockerfile
JingJZ160: Yeah I'll remove them.

docker/Dockerfile
JingJZ160: Will do.

scripts/pull_salesforce.py
JingJZ160: That's my question too. Since we are using GCP, we probably need to setup GCP secret manager as we can't check in those files here. That is going to be a separate PR.

PR #72, opened 2020-05-13
Run submit-to-GCP-training job on Celery

edu-gp:
1. The submit-to-GCP-training job used to run just on the web server. It's now moved to Celery.
2. I added the following pattern to all the Celery jobs. It fixes a minor error, just for correctness.
```
db = Database(...)
try:
     ...
finally:
     db.session.close()
```

### DIFF
--- a/train/train_celery.py
+++ b/train/train_celery.py
@@ -6,6 +6,8 @@ from train.no_deps.run import (
     train_model as _train_model,
     inference as _inference
 )
+from train.gcp_job import GCPJob
+from train.gcp_celery import poll_status as gcp_poll_status
 
 app = Celery(
     # module name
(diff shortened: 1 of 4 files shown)


train/train_celery.py
```diff
@@ -26,24 +28,51 @@
...
+
+        model = prepare_task_for_training(db.session, task.id)
+
+        files_for_inference = task.get_data_filenames(abs=True)
+
+        job = GCPJob(model.uuid, model.version)
+
+        # TODO: A duplicate job would error out. Currently errors for
+        # background jobs are silent...
+        job.submit(files_for_inference)
```
JingJZ160, line 71: Hmm duplicate or not, do we have a way to catch errors?

edu-gp: For _all_ the celery jobs we currently don't have a good way to surface errors. Definitely something we should address. Would you like to take this on? I have some draft code to get started.

PR #75, opened 2020-05-13
Move patterns from Tasks to labels via a LabelPatterns table

edu-gp:
Instead of a Task having multiple patterns, now a "label" has multiple patterns.

We need this change to support multi-label tasks.

This change:
1. Creates a `LabelPatterns` object to hold patterns for labels.
2. User can edit and view patterns for the label.
3. Disable editing patterns on the task object.

I didn't delete patterns from the Task object yet. A future task is to migrate existing labels in Task to LabelPatterns.

This is also not hooked up inside the "Request more Annotations" flow yet. That's also a future PR.


backend/labels.py
```diff
@@ -0,0 +1,89 @@
...
+        guide = get_or_create(db.session, AnnotationGuide, label=label)
+        guide.set_text(annotation_guide_text)
+        db.session.add(guide)
+
+        label_patterns = get_or_create(
+            db.session, LabelPatterns, label=label)
+        label_patterns.set_positive_patterns(patterns)
+        db.session.add(label_patterns)
+
+        db.session.commit()
```
JingJZ160, line 71: Should we add a try-except for this part in case we need to rollback?

PR #78, opened 2020-05-14
Use label patterns in AR

edu-gp:
Use patterns on each label (stored in LabelPatterns) to construct the model used in generating AnnotationRequest's.


ar/__init__.py
```diff
@@ -23,7 +24,24 @@
...
+    if label_patterns:
+        patterns = label_patterns.get_positive_patterns()
+        if len(patterns) > 0:
+            patterns = _convert_to_spacy_patterns(patterns)
+            model = PatternModel(patterns)
+            return model
+        else:
+            return None
+    else:
+        return None
```
JingJZ160, line 41: It could be just one `return None` at the end of the function.

PR #79, opened 2020-05-14
Script to migrate patterns from Task to LabelPatterns

edu-gp:
Loads in pattern and pattern_file from Task, and write those into LabelPatterns.

Will deprecate pattern and pattern_file from Task in a future PR.

### DIFF
--- a/scripts/label_patterns_migration.py
+++ b/scripts/label_patterns_migration.py
@@ -1,0 +1,48 @@
+"""
+Patterns used to be stored on Tasks.
+This script migrates all the patterns from binary-classification tasks
+to individual labels (by storing them in LabelPatterns).
+"""
+from db.model import (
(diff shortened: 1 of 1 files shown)


scripts/label_patterns_migration.py
```diff
@@ -0,0 +1,48 @@
...
+            pat.set_positive_patterns(patterns)
+            db.session.add(pat)
+            db.session.commit()
+
+            print(f'Now there are {pat.count()} patterns for label={label}')
+            print(pat.get_positive_patterns())
+        else:
+            print(f'** Skip task name="{task.name}" id={task.id}, '
+                  f'it has {len(labels)} labels.')
+        print('-'*80)
```
JingJZ160, line 48: Do we still need this line?

edu-gp: Yea, makes it easier to read the output

PR #82, opened 2020-05-14
Train model for a label, and UI changes

edu-gp:
- Hook up model training for each label (in backend/tasks)
- Script to associate Models with labels.
- UI touchups.

### DIFF
--- a/backend/templates/tasks/show.html
+++ b/backend/templates/tasks/show.html
@@ -311,10 +311,7 @@
                   <small class='text-muted ml-3'>{{mv.uuid}}</small>
                 </h5>
 
-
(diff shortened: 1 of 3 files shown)


backend/templates/tasks/show.html
```diff
@@ -242,227 +221,243 @@ <h6>
...
+              <div class="card-body">
+                <h5>
+                  Version {{mv.version}}
+                  <small class='text-muted ml-3'>{{mv.uuid}}</small>
+                </h5>
+
+
+                {% set metrics = mv.get_metrics() %}
+
+                {% if metrics %}
```
edu-gp, line 317: use mv.is_read() instead

PR #88, opened 2020-05-25
Using pattern model for each label to create pattern matches. Fix bug

JingJZ160:
- Use pattern model for labels
- Merge pattern models from different labels together. For the same entity, the tokens are the same. All matches are grouped together and sorted (easier for unit test). The scores are also saved although I don't know how we are going to display the scores for multiple labels.

### DIFF
--- a/ar/__init__.py
+++ b/ar/__init__.py
@@ -197,9 +197,25 @@ def generate_annotation_requests(dbsession, task_id: int,
         __text_list.append(row['text'])
 
     # Pattern decorations
+    __pattern_decor_for_all_labels = []
     if task.get_pattern_model():
-        __pattern_decor = task.get_pattern_model().predict(__text_list,
-                                                           fancy=True)
+        labels = task.get_labels()  # ["hotdog", "fake"]
+        logging.error(labels)
+        for label in labels:
+            pattern_model_for_label = get_pattern_model_for_label(
+                dbsession=dbsession, label=label)
+            logging.error(pattern_model_for_label)
+            if pattern_model_for_label:
+                __pattern_decor_for_label = pattern_model_for_label.predict(
+                    __text_list, fancy=True
+                )
+                __pattern_decor_for_all_labels.append(__pattern_decor_for_label)
+
+        __pattern_decor = _merge_pattern_decor_for_all_labels_recursive(
+            pattern_decor_for_all_labels=__pattern_decor_for_all_labels,
+            low=0,
+            high=len(__pattern_decor_for_all_labels) - 1
+        )
     else:
         __pattern_decor = None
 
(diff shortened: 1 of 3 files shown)


ar/__init__.py
```diff
@@ -197,9 +197,25 @@ def generate_annotation_requests(dbsession, task_id: int,
         __text_list.append(row['text'])
 
     # Pattern decorations
+    __pattern_decor_for_all_labels = []
     if task.get_pattern_model():
-        __pattern_decor = task.get_pattern_model().predict(__text_list,
-                                                           fancy=True)
+        labels = task.get_labels()  # ["hotdog", "fake"]
```
edu-gp, line 202: Remove comment please

ar/__init__.py
```diff
@@ -197,9 +197,25 @@ def generate_annotation_requests(dbsession, task_id: int,
...
+        logging.error(labels)
```
edu-gp, line 203: Remove debug message

ar/__init__.py
```diff
@@ -197,9 +197,25 @@ def generate_annotation_requests(dbsession, task_id: int,
         __text_list.append(row['text'])
 
     # Pattern decorations
+    __pattern_decor_for_all_labels = []
     if task.get_pattern_model():
```
edu-gp, line 201: `task.get_pattern_model()` is no longer relevant...

ar/__init__.py
```diff
@@ -197,9 +197,25 @@ def generate_annotation_requests(dbsession, task_id: int,
...
-        __pattern_decor = task.get_pattern_model().predict(__text_list,
-                                                           fancy=True)
+        labels = task.get_labels()  # ["hotdog", "fake"]
+        logging.error(labels)
+        for label in labels:
+            pattern_model_for_label = get_pattern_model_for_label(
+                dbsession=dbsession, label=label)
+            logging.error(pattern_model_for_label)
```
edu-gp, line 207: Remove debug message

ar/__init__.py
```diff
@@ -223,6 +239,49 @@ def generate_annotation_requests(dbsession, task_id: int,
     return annotation_requests
 
 
+def _merge_pattern_decor_for_all_labels_recursive(
+        pattern_decor_for_all_labels: List[List], low: int, high: int):
```
edu-gp, line 243: Can you add some documentation?

And sorry can you also explain to me why we need to do (what looks to me like) a merge sort here?

ar/__init__.py
JingJZ160: Done.

ar/__init__.py
JingJZ160: Done.

ar/__init__.py
JingJZ160: Done.

ar/__init__.py
JingJZ160: Done.

ar/__init__.py
JingJZ160: As discussed, we are going to keep this logic. The documentation is added to the code.

JingJZ160:
Local test was done by adding a fake pattern model for label "fake". The patterns includes "the" and "cloud". The other label is "hotdog" but the patterns are from the healthcare patterns.

Just to show that it works:
image

image

PR #89, opened 2020-05-26
Admin correction workflow

JingJZ160:
This PR is on top of the ["Show Existing Annotation" PR](https://github.com/edu-gp/annotation_tool/pull/87). Changes include:

- A table in the backend server webpage to show the links of examining all users' existing annotations.
- An `examine` API in the frontend server to process the workflow of examining users' annotations.
- The `examine` reuses the `reannotate` and `update_annotation` APIs to display and update existing annotations with two additional parameters `is_admin_correction` and `username_under_exam`. They help to distinguish between a normal user correcting his/her own annotations and an admin examining all annotations from a user.

I'm assuming only the admin have access to the backend server page to see the `examine` links. This is of course a very naive assumption. If we want to have user-role management, we probably need a separate task.

If we are OK with everyone correcting each other's annotations, then we don't need the role management part. As long as they can get the examination link by accessing the backend server (or some other ways), they can start the workflow.

Backend:
```
Admin Examine Links

    User1: http://localhost:5001/tasks/6/examine/user1
    User2: http://localhost:5001/tasks/6/examine/user2
    User3: http://localhost:5001/tasks/6/examine/user3
```
Frontend:
```
hotdog
Existing Annotations from user1 
Entity | Example
-- | --
1 | Example syapse.com
2 | Example mammoth.bio
3 | Example mpulsemobile.com
4 | Example healthiq.com
```
By using the link on the backend, they can get to the page of annotations done by a user for the labels under a task. **Note that if there are annotations coming from other sources by the same user but have the same label, they will show up here too. This is why we may want to add a source column to the `Annotation` table since `Task` doesn't have direct link to `Annotation`.** 

Each time an annotation is examined, it will move to the next one in the list. When it's done, it will return to the list page.


edu-gp:
OK with everyone correcting each other's annotations; No need for role management just yet.

backend/templates/tasks/show.html
```diff
@@ -184,6 +184,19 @@ <h6>Annotator Login Links</h6>
           </div>
         </div>
 
+      <div class="row">
+          <div class="col">
+            <h6>Admin Examine Links</h6>
+            <ul>
+              {% for user, link in admin_examine_links %}
+              <li>
+                <b>{{user}}</b>: <span onclick="selectText(this.id)" id="{{link}}">{{link}}</span>
```
edu-gp, line 193: Let's just make this a `<a href="{{link}}" target="_blank">{{user}}</a>`

db/model.py
```diff
@@ -765,8 +765,11 @@ def _raw_data_file_path(fname):
     return os.path.join(filestore_base_dir(), RAW_DATA_DIR, fname)
 
 
-def fetch_annotation_entity_and_ids_done_by_user_under_task(dbsession, username,
-                                                            labels):
+# TODO this does not guarantee to fetch annotations under a task. It only
+#  fetch annotations with the labels under a task. If we are looking for
+#  annotations created inside Alchemy website, we'd better add a source column.
```
edu-gp, line 770: True, and to your point a "source" column would address this. How about we remove this TODO here and turn it into a task on clickup?

frontend/static/annotation_box.js
```diff
@@ -187,6 +187,8 @@ class AnnotationBox extends React.Component {
             var yes_btn_icon = String.fromCharCode(10003); // check
             var no_btn_icon = String.fromCharCode(10007); // cross
 
+            console.log(anno['labels'])
```
edu-gp, line 190: Debug code

frontend/tasks.py
```diff
@@ -176,13 +211,22 @@ def update_annotation():
...
+        else:
+            return {'redirect': url_for('tasks.reannotate', task_id=task_id,
+                                        annotation_id=next_annotation_id)}
     else:
-        return {'redirect': url_for('tasks.show', id=task_id)}
+        if is_admin_correction:
+            return {'redirect': url_for('tasks.examine', task_id=task_id,
+                                        user_under_exam=username_under_exam)}
+        else:
+            return {'redirect': url_for('tasks.show', id=task_id)}
```
edu-gp, line 229: Instead of passing in `is_admin_correction` and `username_under_exam`, how about just pass in the `redirect` URL? That's the common practice, and makes this function more reusable.

I would say the functionality to go to "Next" in the review is optional... if it makes the code unclean then let's just rip it out.

backend/templates/tasks/show.html
JingJZ160: Done.

db/model.py
JingJZ160: Click up task created: https://app.clickup.com/t/7eexk7

frontend/static/annotation_box.js
JingJZ160: Removed.

frontend/tasks.py
JingJZ160: OK I'll remove the go to Next functionality and pass in `redirect_link` instead.

JingJZ160:
ping :)

PR #93, opened 2020-06-01
Bulk Annotate

edu-gp:
Introduce a section underneath each Annotations area with links for bulk uploading annotations.

Screen Shot 2020-06-02 at 11 39 25 AM

This redirects to a new page where users can insert new annotations like this:

Screen Shot 2020-06-01 at 4 26 18 PM


backend/templates/tasks/show.html
```diff
@@ -281,49 +281,65 @@ <h5>Annotations by Labels</h5>
...
+          <td class="align-middle">
+            <a class="btn btn-secondary btn-sm" href="{{url_for('annotations.bulk', user=user, label=label)}}"
+              target="_blank">{{label}}</a>
+          </td>
+          {% endfor %}
+        </tr>
+        {% endfor %}
+      </tbody>
+    </table>
+    <a class="btn btn-secondary btn-sm" href="{{url_for('annotations.bulk')}}" target="_blank">Bulk Upload</a>
```
JingJZ160, line 342: Why do we need a table for per label per user bulk upload and another link for bulk upload in general?

Do we also need a table head?

edu-gp: It looks weird with a table head actually...

I figured let's just throw all the links there (the bulk upload per label and per user auto-populates the label and user fields, so it's less prone to errors)

Can definitely change the UI in the future!

PR #94, opened 2020-06-01
bug fix for not updating the annotations with multiple labels

JingJZ160:
Bug fix for not being able to update annotations with multiple labels plus a fix for a navigation issue. It's strange the old code actually worked locally...

### DIFF
--- a/frontend/tasks.py
+++ b/frontend/tasks.py
@@ -243,6 +243,7 @@ def update_annotation():
         ).one_or_none()
         if annotation:
             annotation.value = value
+            annotation.context = context
         else:
             # In case the user forgot to annotate this label in the last pass.
             annotation = ClassificationAnnotation(
--- a/frontend/tasks.py
+++ b/frontend/tasks.py
@@ -15,7 +15,7 @@ from flask import (
 from db.model import (
     db, AnnotationRequest, Task, get_or_create, ClassificationAnnotation,
     AnnotationGuide, AnnotationValue, AnnotationRequestStatus,
-    fetch_annotation_entity_and_ids_done_by_user_under_labels)
+    fetch_annotation_entity_and_ids_done_by_user_under_labels, User)
 
 from .auth import login_required
 
@@ -115,7 +115,7 @@ def annotate(task_id, ar_id):
         username=g.user['username']
     )
 
-    anno["update_redirect_link"] = request.referrer
+    anno["update_redirect_link"] = url_for('tasks.show', id=task_id)
     anno["task_page_name"] = "Task"
 
     return render_template('tasks/annotate.html',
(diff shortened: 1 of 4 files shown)


scripts/update_existing_annotation_context.py
```diff
@@ -19,13 +19,15 @@
...
     for annotation in annotations:
         if annotation.entity in entity_to_context:
             logging.info("Updating Entity {}".format(annotation.entity))
             annotation.context = entity_to_context[annotation.entity]
+            to_update.append(annotation)
         else:
             logging.info("Did not find request for entity {}".format(annotation.entity))
 
-    db.session.add_all(annotations)
+    db.session.add_all(to_update)
```
edu-gp, line 31: Is it safe to rerun this script?

frontend/tasks.py
```diff
@@ -219,16 +219,40 @@ def update_annotation():
...
-                                   id=annotation_id)
-        annotation.value = value
+        annotation = db.session.query(ClassificationAnnotation).filter(
+            ClassificationAnnotation.label == label,
+            ClassificationAnnotation.entity == entity,
+            ClassificationAnnotation.entity_type == entity_type,
+            ClassificationAnnotation.user_id == annotation_owner.id
+        ).one_or_none()
+        if annotation:
+            annotation.value = value
```
edu-gp, line 245: Probably should assign the latest `context` as well.

frontend/tasks.py
```diff
@@ -219,16 +219,40 @@ def update_annotation():
...
+        else:
+            # In case the user forgot to annotate this label in the last pass.
+            annotation = ClassificationAnnotation(
+                label=label,
+                entity_type=entity_type,
+                entity=entity,
+                user_id=annotation_owner.id,
+                value=value,
+                context=context
+            )
```
edu-gp, line 255: So, previously, a user could've had multiple annotations for the same entity, right? Wondering if we should write a script to clean that up.

For the future: It would be useful to see the history of user annotations on an entity - We could store all user actions in an audit trail, perhaps?

scripts/update_existing_annotation_context.py
JingJZ160: Yes.

frontend/tasks.py
JingJZ160: Sure I'll add it to make it safer.

frontend/tasks.py
JingJZ160: Do you mean multiple annotations for different labels or the same label? If the former, do we need to clean them up? If for the same label, then yes.

JingJZ160: For the audit trail, we can add a version column to the annotations but what is the use case for seeing the history? Educational purpose?

edu-gp: For the same label - Okay yea, a task for the future, to look through existing db to clean it up. I don't think we enforced that originally.

Yea for educational purposes, where people tend to disagree, and the performance of an annotator over time. If we keep it within the Annotation table, then 1) we can't enforce uniqueness on (user, entity, entity_type) and 2) whenever we want to query the annotations, we need to add a query on fetching just the latest annotation grouped by (user, entity, entity_type). So I figured it might be cleaner to keep it as a separate table (we could also keep it in the logs, that's another option)

scripts/update_existing_annotation_context.py
edu-gp: Okay, please remind me to run it in prod if I forget!

edu-gp:
LGTM

frontend/tasks.py
JingJZ160: https://app.clickup.com/t/7mjfqf

https://app.clickup.com/t/7mcf18

PR #95, opened 2020-06-02
Adding progress display on the annotate page for requests

JingJZ160:
Progress bar for annotating requests. It seems we need to pass around the list id and total number of requests in the URL. Otherwise we may need to write queries for every page.

### DIFF
--- a/frontend/tasks.py
+++ b/frontend/tasks.py
@@ -191,7 +191,7 @@ def receive_annotation():
 
     if next_ar_id:
         return {'redirect': url_for('tasks.annotate', task_id=task_id,
-                                    ar_id=next_ar_id, item_id=item_id)}
+                                    ar_id=next_ar_id)}
     else:
         return {'redirect': url_for('tasks.show', id=task_id)}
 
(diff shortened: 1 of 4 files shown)


frontend/tasks.py
```diff
@@ -108,16 +108,22 @@ def compare_annotations(task_id):
...
         example_id=ar_id,
         is_request=True,
-        username=g.user['username']
+        username=annotator
     )
 
     anno["update_redirect_link"] = url_for('tasks.show', id=task_id)
     anno["task_page_name"] = "Task"
 
+    anno["item_id"] = request.args.get("item_id", None)
```
edu-gp, line 122: Is `item_id` the AnnotationRequest id?

If we want to show progress, shouldn't we just show how many annotation requests the user has completed? (A SQL query)

JingJZ160: No, `item_id` is just the item index of the task list on the `show` page. I'm thinking this is to show which annotation out of the 50 in the list is displayed right now and that represents the progress. But I guess your definition is more legit.

PR #96, opened 2020-06-02
Download predictions with each user annotations + stats

edu-gp:
When the users download predictions right now, they get:

```
domain, text, prob
a.com , "..", 0.99
b.com , "..", 0.86
```

Users have requested to also include 1) all the annotations from the users and 2) contention score + majority vote. For example:

```
domain, text, prob, annotator_a, annotator_b, annotator_c, contention, majority_vote
a.com , "..", 0.99,           1,           0,           1,          0,             1
b.com , "..", 0.86,          -1,          -1,           1,        0.6,            -1
```

This PR lumps all the code to do so in one function - it can use some refactoring later, but I think this is the entire functionality.

### DIFF
--- a/backend/tasks.py
+++ b/backend/tasks.py
@@ -1,5 +1,5 @@
 import logging
-from typing import List
+from typing import List, Optional
 
 from flask import (
     Blueprint, flash, redirect, render_template, request, url_for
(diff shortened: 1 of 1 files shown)


backend/tasks.py
```diff
@@ -244,10 +244,83 @@ def download_prediction():
...
+            entropy = -sum([p*logp for p, logp in zip(probs, log_probs)])
+            return entropy
+
+        def _get_majority_vote(annos: List[Optional[int]]):
+            cnt = _build_counter(annos)
+
+            if len(cnt):
+                return cnt.most_common(1)[0][0]
+            else:
+                return None
```
JingJZ160, line 316: Can you extract these functions to a util module? THe contentious level calculation I used was pretty naive. We should probably standardize it by using the same formula.

edu-gp: Okay, let me do that in a separate PR so we can get this one deployed

PR #97, opened 2020-06-02
Use ENV variable to replace the db_url in alembic ini

JingJZ160:
Use environment variable to replace the db_url in the alembic ini file. This is only used for DB migration purpose. Locally it works but we will need to test it on the prod server with the real DB to verify if it works there.


JingJZ160:
> LGTM. Can you keep a note somewhere so we don't forget to test this when we deploy?

Created a task and put in selected section.

PR #98, opened 2020-06-03
Adding try except block vs tear_down_request

JingJZ160:
We can either wrap try-except on every commit or use the `tear_down_request` to handle the commit and rollback (see the commented-out code block). 

The tear_down_request will be executed after each request even if the request failed. My only concern is that it makes me uncomfortable not seeing try-catch block after we use `db.session.add` in the controller code. I have to mentally recall the `tear_down_request` is handling it. What's your thought on this?

### DIFF
--- a/frontend/__init__.py
+++ b/frontend/__init__.py
@@ -46,29 +46,14 @@ def create_app(test_config=None):
     @login_required
     def index():
         username = g.user['username']
-        task_ids = fetch_tasks_for_user(username)
         task_id_and_name_pairs = fetch_tasks_for_user_from_db(
             db.session, username)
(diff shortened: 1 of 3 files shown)


frontend/__init__.py
```diff
@@ -56,6 +57,19 @@ def index():
...
+    #         logging.error(error)
+    #         db.session.rollback()
+    #         raise error
+    #     try:
+    #         db.session.commit()
+    #     except DatabaseError as e:
+    #         logging.error(e)
+    #         db.session.rollback()
+    #         raise e
+
```
edu-gp, line 72: Let's just remove this

JingJZ160: Done.

PR #99, opened 2020-06-04
Remove requests under deleted annotator during task edit.

JingJZ160:
Verified by 
- Check DB that there are requests under task T1 with user_id U1
- Delete user U1 from the task T1
- Check requests under task T1 for U1 are gone
- Request more annotations
- No requests for U1 under task T1 are created

Questions: does it make sense to do the same when removing labels (and changing data file)?

### DIFF
--- a/backend/tasks.py
+++ b/backend/tasks.py
@@ -4,8 +4,11 @@ from typing import List, Optional
 from flask import (
     Blueprint, flash, redirect, render_template, request, url_for
 )
+from sqlalchemy.exc import DatabaseError
+from werkzeug.urls import url_encode
 
-from db.model import db, Task, Model, AnnotationGuide, LabelPatterns
+from db.model import db, Task, Model, AnnotationGuide, LabelPatterns, \
+    AnnotationRequest, User, delete_requests_for_user_under_task
 from db.utils import get_all_data_files
 from ar.data import compute_annotation_statistics_db, \
     compute_annotation_request_statistics
@@ -188,10 +191,21 @@ def update(id):
         task.name = name
         task.set_data_filenames([data])
         task.set_labels(labels)
+
+        for current_annotator in task.get_annotators():
+            if current_annotator not in annotators:
+                logging.info("Prepare to remove requests under user {} for "
+                              "task {}".format(current_annotator, id))
+                delete_requests_for_user_under_task(db.session,
+                                                    current_annotator,
+                                                    id)
         task.set_annotators(annotators)
 
         db.session.add(task)
+
         db.session.commit()
+        logging.info("Updated tasks and deleted requests from removed "
+                      "annotators.")
         return redirect(url_for('tasks.show', id=task.id))
     except Exception as e:
         db.session.rollback()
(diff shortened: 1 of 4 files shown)


backend/tasks.py
```diff
@@ -195,11 +197,30 @@ def update(id):
                                list_to_textarea=list_to_textarea)
     else:
         task.name = name
+        # TODO should we also remove requests when user update the data
+        #  file and label?
```
edu-gp, line 201: +1

backend/tasks.py
```diff
@@ -186,12 +190,38 @@ def update(id):
...
+                    delete_requests_for_user_under_task(db.session,
+                                                        current_annotator,
+                                                        id)
+            # Updating the labels
+            for current_label in task.get_labels():
+                if current_label not in labels:
+                    logging.info("Prepare to remove requests under label {} for "
+                                 "task {}".format(current_label, id))
+                    delete_requests_for_label_under_task(db.session,
+                                                         current_label, id)
```
edu-gp, line 214: Would it be cleaner if the deletion of AnnotationRequest can be triggered from the Model? E.g. if we can add a listener to a field in the Task, if a specific field is edited & saved, SQLAlchemy knows to remove requests as needed? I'm not sure if this overkill. What do you think?

db/model.py
```diff
@@ -783,3 +783,29 @@ def fetch_annotation_entity_and_ids_done_by_user_under_labels(
...
+    dbsession.query(AnnotationRequest). \
+        filter(AnnotationRequest.task_id == task_id,
+               AnnotationRequest.label == label). \
+        delete(synchronize_session=False)
+
+
+def delete_requests_under_task(dbsession, task_id):
+    dbsession.query(AnnotationRequest). \
+        filter(AnnotationRequest.task_id == task_id). \
+        delete(synchronize_session=False)
```
edu-gp, line 811: Can we have some unit tests for these db functions?

db/model.py
```diff
@@ -783,3 +783,29 @@ def fetch_annotation_entity_and_ids_done_by_user_under_labels(
...
     return res
+
+
+def delete_requests_for_user_under_task(dbsession, username, task_id):
+    # Can't call Query.update() or Query.delete() when join(),
+    # outerjoin(), select_from(), or from_self() has been called. So we have
+    # to get the user instance first.
+    user = get_or_create(dbsession=dbsession,
+                         model=User,
+                         username=username)
```
edu-gp, line 794: Why `get_or_create` and not just a "get"?

Also, this is assuming user is always not null. Let's add a check for that.

backend/tasks.py
JingJZ160: Done.

db/model.py
JingJZ160: Done.

db/model.py
JingJZ160: Done.

backend/tasks.py
JingJZ160: You mean we should put the logic in the `setter` method of the `Task` class in the `model.py`? I actually prefer not doing that. I'd rather have something explicit than adding extra meaning in a `setter` method. If this makes the code in `update` method to clunky, I can move them to a helper method.

PR #100, opened 2020-06-04
Adding the source column with a one-time backfill script

jzhang-gp:
Adding the source column. It may take two steps and this is the first one. For now I need to set source as nullable=True. Once we backfill the data, we can change it to be nullable=False. 

TODO: need to change some of the queries on existing annotations done by users on the website. Adding the source column as a filtering criterion.


jzhang-gp:
As discussed, we are not going to add this column

PR #101, opened 2020-06-09
Script to update data source

JingJZ160:
Temporary script for updating the source data. I'll probably move it to a separate docker image once we are ok with the logic here.

All key and bucket names are stored in secret manager.


JingJZ160:
Moving to a different repo. Closing.

PR #102, opened 2020-06-09
First pass on inference on new data

edu-gp:
The algorithm to do this is:

1. Check if the model has made an inference on any files previously. If yes, use the predictions on those files to build an in-memory cache.
2. For each row in the file we're doing inference on, collect all the rows that are not in cache yet. Send only these rows to the model for inference.
3. Merge the inference results with the cached results, store that as the final inference result for the new file

Also includes some refactoring & tests.

TODOs (probably in future PRs):

- [x] A loop to run through all models on the new data.
- [x] Interface with the rest of Alchemy.
- [x] Test this locally end-to-end.
- [x] Rebuild the training docker image, update it on prod server.
- [x] Test this on Google AI Platform

### DIFF
--- a/db/utils.py
+++ b/db/utils.py
@@ -33,3 +33,7 @@ def get_all_pattern_files():
     d = os.path.join(filestore_base_dir(), RAW_DATA_DIR)
     return sorted([x for x in os.listdir(d)
                    if is_pattern_file(os.path.join(d, x))])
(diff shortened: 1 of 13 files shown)


edu-gp:
Tested with local data, with an existing dataset and an example new dataset.

PR #103, opened 2020-06-16
Adding logging to the website

jzhang-gp:
Use GCP cloud logging to emit logs to the Logging service. 

Only issue right now is that somehow the celery logs are not emitted


backend/__init__.py
```diff
@@ -1,13 +1,24 @@
...
+
 from db.config import DevelopmentConfig
 from db.model import db
 from .auth import auth
 
+from google.cloud.logging.handlers import CloudLoggingHandler, setup_logging
+
+client = glog.Client()
+
+handler = CloudLoggingHandler(client, name="alchemy-backend")
```
edu-gp, line 18: Can we extract the name to an env var?

docker/Dockerfile
```diff
@@ -9,6 +9,15 @@ RUN pip install mockito flask-sqlalchemy alembic Flask-Testing psycopg2-binary
...
 RUN pip install google-cloud-secret-manager simple-salesforce PyJWT cryptography requests tldextract
 
+# GCP Cloud logging
+RUN pip install google-cloud-logging
+
+# AWS CLI
+RUN pip install boto3
+
+# Spring3 data fetcher
+RUN pip install ijson tqdm google-cloud-storage
```
edu-gp, line 19: Can we move the deps for AWS to another change? Doesn't look like they belong here.

frontend/__init__.py
```diff
@@ -6,13 +6,23 @@
...
 from .auth import login_required
 
 from db._task import _Task
 from ar.data import fetch_tasks_for_user, fetch_tasks_for_user_from_db
 
+from google.cloud.logging.handlers import CloudLoggingHandler, setup_logging
+
+client = glog.Client()
+
+handler = CloudLoggingHandler(client, name="alchemy-frontend")
```
edu-gp, line 22: Ditto here, extract to env var

jzhang-gp: Done.

docker/Dockerfile
jzhang-gp: Removed.

backend/__init__.py
jzhang-gp: Done.

PR #104, opened 2020-06-16
Add a radio button to set active model 

jzhang-gp:
Add a primitive form with radio buttons in the model section. The UI is a bit ugly but we can make it look better later.

The functionality include selecting an active, updating the `is_active` column of a model instance and redirecting back to the task show page.

Follow-up question: where are we using the active model instance? The functions like `get_active_model` or `get_active_nlp_model` are not in effect.


db/model.py
```diff
@@ -316,6 +317,7 @@ class Model(Base):
     # training system. UUID makes sure those jobs don't clash.
     uuid = Column(String(64), index=True, nullable=False, default=gen_uuid)
     version = Column(Integer, index=True, nullable=False, default=1)
+    is_active = Column(Boolean, default=False)
```
edu-gp, line 320: Is it possible to enforce at most 1 active Model per Label?

backend/templates/tasks/show.html
```diff
@@ -367,8 +367,19 @@ <h4>
...
-
+            <form action="{{ url_for('models.update_active_model') }}" method="POST">
+                <button type="submit" class="btn btn-primary">Update Active Model</button>
             {% for mv in models_per_label.get(label, []) %}
+                <div class="card mt-4">
+                 <div class="card-body">
+                     {% if mv.is_active %}
+                        <input type="radio" name="model_id" value="{{mv.id}}" checked="checked"/> Active Model
+                     {% else %}
+                        <input type="radio" name="model_id" value="{{mv.id}}"/> Set Active
```
edu-gp, line 378: Is it possible to enforce at most 1 active Model per Label is the UI?

db/model.py
```diff
@@ -564,6 +566,9 @@ def get_latest_model(self):
 
     def get_active_nlp_model(self):
         from inference.nlp_model import NLPModel
+        # TODO change this to be getting the active model specified by the
+        #  user. We can default to the latest version but should allow user
+        #  to choose their own.
```
edu-gp, line 571: Can we include that change here as well?

backend/models.py
```diff
@@ -48,3 +51,37 @@ def export_new_raw_data():
...
+def update_active_model():
+    model_id = request.form.get("model_id")
+    new_active_model = db.session.query(Model).\
+        filter(Model.id == model_id).one_or_none()
+    logging.error(new_active_model)
+
+    current_active_model = db.session.query(Model).\
+        filter(Model.label == new_active_model.label,
+               Model.type == new_active_model.type,
+               Model.task_id == new_active_model.task_id,
```
edu-gp, line 66: Don't have to check task_id; just label and type are fine.

backend/models.py
```diff
@@ -1,6 +1,9 @@
+import logging
+
 from flask import (
-    Blueprint, request, jsonify
-)
+    Blueprint, request, jsonify,
+    redirect)
```
edu-gp, line 5: nit: can all fit on the same line

edu-gp:
To your question in the description: We use active model when 1) exporting data and 2) active learning when generating annotation requests.

backend/templates/tasks/show.html
jzhang-gp: Doesn't radio button natively handles that? It doesn't allow more than 1 item to be selected.

backend/models.py
jzhang-gp: Done.

backend/templates/tasks/show.html
edu-gp: Is it all radio buttons in the same form, or all radio buttons on the page?

If the latter, then if there are multiple Labels on the same page (for multi-label tasks), would their radio buttons conflict with each other?

jzhang-gp: For the UI, under each label there is a separate form with multiple radio buttons. So if there are two different labels in a task, it will include two forms and they will not conflict each other.

db/model.py
jzhang-gp: Hmm I didn't find any native constraint within SQLAlchemy. We might have to write something ourselves if want this check. For now, we are using `one_or_none` in the query call so if there are multiple active models for the same label, it will throw an exception. 

(Not ideal solution) I can move the logic in the `update_active_model` to the `Model` class so it's easier to add unit test to make sure only 1 active model is present. We can enforce to use this function in all our future code.

db/model.py
jzhang-gp: I've checked the repo and found no reference to this method. 

When generating annotation requests, we are using a different method `get_nlp_model_for_label`. 

When we are exporting the data for a model, we are using the model that the user clicks on. It could be either active or inactive. 

So I think we can just get rid of this method.

backend/models.py
jzhang-gp: So do we still the task id on the model object? We are assigning that field when `prepare_next_model_for_label`. Older models on my local db have that field and without it in the query, it will cause trouble. 

I wonder if we have similar issue on production db.

backend/templates/tasks/show.html
edu-gp: 👍

db/model.py
edu-gp: Got it, I'm good with removing `get_active_nlp_model` here, in that case can we move `get_nlp_model_for_label` from ar/__init__.py to here? Seems like it belongs in this file with all the other db methods.

backend/models.py
edu-gp: Agreed, I think it's cleaner to remove task_id from Model (and the "models" relationship from Task).

So the clean setup looks to me like:

Task <--many_to_many--> Label <--has_many--> Model

db/model.py
edu-gp: One way is to store this information in a table like `LabelActiveModel` with 2 columns, `label` and `model_id`, and we can put a unique composite index on both columns. What do you think?

jzhang-gp:
Updated the deployment config changes:

If the model is not ready, it will not show the Approve buttons.
image

When the model is ready (I have to hardcode it to make it work since we have that bug on local training...)
It shows 3 levels of options as discussed:
1. Approve button
2. If approved, we show the deployment radion button
3. If selected, we show the threshold text box

When approval is not checked, the other two will be hidden.
When approved but un-selected for deployment, the threshold box will hide.

This is not the best UI as the options should align with the model version info on the same line but we can work on that later if the rest of the logic is good to go. 
image

edu-gp:
I think some code changes for this are in your other PR https://github.com/edu-gp/annotation_tool/pull/105

PR #105, opened 2020-06-29
Modify entity type

jzhang-gp:
As discussed, we added entity_type as part of the task creation process. Other changes follow that to add `entity_type` in functions that assign `entity_type` to COMPANY. 

The missing part is a script to backfill the `entity_type` to the existing tasks in the db. I'll add it later.


alembic/versions/a94911ed125b_remove_relationship_between_task_and_.py
```diff
@@ -0,0 +1,34 @@
...
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('model', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('task_id', sa.INTEGER(), nullable=True))
+        batch_op.create_foreign_key('fk_model_task_id_task', 'task', ['task_id'], ['id'])
+
+    # ### end Alembic commands ###
```
edu-gp, line 34: You're probably aware but this file is already in https://github.com/edu-gp/annotation_tool/pull/104/files

alembic/versions/e05604e8db05_add_is_active_model_field.py
```diff
@@ -0,0 +1,32 @@
...
+
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('model', schema=None) as batch_op:
+        batch_op.drop_column('is_active')
+
+    # ### end Alembic commands ###
```
edu-gp, line 32: We don't need this anymore right?

db/model.py
```diff
@@ -821,6 +816,31 @@ def delete_requests_for_label_under_task(dbsession, label, task_id):
...
+                                              task_id=task_id,
+                                              entity_type=entity_type)
+
+
 def delete_requests_under_task(dbsession, task_id):
     delete_requests_under_task_with_condition(dbsession,
                                               task_id=task_id)
+
+
+def get_active_model_for_label(dbsession, label,
```
edu-gp, line 831: I'm a bit confused. Is `get_active_model_for_label` suppose to be together with the other PR https://github.com/edu-gp/annotation_tool/pull/104/files ?

alembic/versions/a94911ed125b_remove_relationship_between_task_and_.py
jzhang-gp: This should be OK once we merge the other PR. I'll resolve the conflicts locally first.

alembic/versions/e05604e8db05_add_is_active_model_field.py
jzhang-gp: hmm it is supposed to be with the other PR. It should be OK to merge it since it is not present in the other PR.

db/model.py
jzhang-gp: Yep it should be with the other PR. I'll merge the other first and then come back to resolve this one.

PR #108, opened 2020-07-03
Show items with highest and lowest predicted probs for annotation

edu-gp:
In the active learning routine, we used to only use the highest entropy items (e.g. predictions near 0.5) as data points to request more annotations.

In this change, we also include top items (predictions near 1) and bottom items (predictions near 0). Because 1) if an item is near the top/bottom, but is actually incorrect, then it's a very high signal data point. And 2) it serves as a natural way to check for model quality.

TODOs:
- [x] Check if this works locally


db/model.py
```diff
@@ -843,5 +843,5 @@ def get_latest_model_for_label(dbsession, label,
     return dbsession.query(Model) \
             .filter_by(label=label,
                        type=model_type) \
-            .order_by(Model.version.desc()) \
+            .order_by(Model.version.desc(), Model.created_at.desc()) \
```
edu-gp, line 846: Some older models that got migrated over has a different `uuid` but the same version number, so to be correct we actually need to sort by version and created_at

PR #109, opened 2020-07-03
Move SQLAlchemy test deps into conftest.py

edu-gp:
Just cleaning up an import that's repeated in way too many places


edu-gp:
> So how do the unit test file find the dependencies without this import?

Pytest imports conftest.py first

PR #110, opened 2020-07-03
Inference Endpoint

edu-gp:
Run inference on a data files on selected-to-be-deployed models and outputs the results `predictions.csv` and `metadata.json` in a `/prod` folder in the GCS bucket.

If the model has not ran inference on the given data file already, it will trigger a Google AI Platform job to run the inference first before returning and trying again.

Also exposes an API endpoint `/api/trigger_inference` to do the above (with simple Bearer auth authentication)

---

TODOs

- [ ] More tests
- [x] Try with a trained model + existing inference file.
- [x] Try with a trained model + new inference file.
- [x] Get feedback from Shah on output format
- [ ] Try with an untrained model + new inference file.
- [ ] (Future) Publish a message for when a new prod inference result is ready.
- [ ] (Future) Do not create new deployed file unless it doesn't exist already.
- [ ] (Future) Combine inferences together for cheaper inference (every week)
- [ ] (Future) Make job polling not in a loop, so we can handle multiple tasks at the same time.
- [ ] (Future) Integrate Flower so we can monitor all the jobs.


db/model.py
```diff
@@ -775,17 +782,17 @@ def _raw_data_file_path(fname):
...
+        ClassificationAnnotation.id,
+        ClassificationAnnotation.created_at,
+        ClassificationAnnotation.label,
+        ClassificationAnnotation.value).join(User). \
+        filter(
+        User.username == username,
+        ClassificationAnnotation.label.in_(labels),
+        ClassificationAnnotation.value != AnnotationValue.NOT_ANNOTATED). \
+        order_by(ClassificationAnnotation.created_at.desc()). \
+        all()
```
edu-gp, line 795: Linter auto-formatted

db/model.py
```diff
@@ -810,7 +817,8 @@ def delete_requests_under_task_with_condition(dbsession, task_id,
 
 
 def delete_requests_for_user_under_task(dbsession, username, task_id):
-    user = dbsession.query(User).filter(User.username == username).one_or_none()
+    user = dbsession.query(User).filter(
+        User.username == username).one_or_none()
```
edu-gp, line 821: Linter auto-formatted

db/model.py
```diff
@@ -841,7 +849,7 @@ def get_latest_model_for_label(dbsession, label,
...
 
     return dbsession.query(Model) \
-            .filter_by(label=label,
-                       type=model_type) \
-            .order_by(Model.version.desc(), Model.created_at.desc()) \
-            .first()
+        .filter_by(label=label,
+                   type=model_type) \
+        .order_by(Model.version.desc(), Model.created_at.desc()) \
+        .first()
```
edu-gp, line 855: Linter auto-formatted

tests/integration/test_api.py
```diff
@@ -0,0 +1,63 @@
+from tests.fixtures import *
+from backend import api
+
+
+def test_hc(backend_client):
+    response = backend_client.get('/api/hc')
```
jzhang-gp, line 6: Is there a file for the fixtures? I can't locate the actual `backend_client` variable...

edu-gp: It's a part of `tests.fixtures` imported via `from tests.fixtures import *`

PR #112, opened 2020-07-03
Refactor functions to calculate entropy and majority vote into shared utils

edu-gp:
As requested in https://github.com/edu-gp/annotation_tool/pull/96

### DIFF
--- a/backend/tasks.py
+++ b/backend/tasks.py
@@ -1,36 +1,42 @@
+import os
 import logging
+import tempfile
+import pandas as pd
 
 from flask import (
-    Blueprint, flash, redirect, render_template, request, url_for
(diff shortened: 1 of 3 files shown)


backend/tasks.py
```diff
@@ -334,40 +335,12 @@ def download_prediction():
...
-            return entropy
-
-        def _get_majority_vote(annos: List[Optional[int]]):
-            cnt = _build_counter(annos)
-
-            if len(cnt):
-                return cnt.most_common(1)[0][0]
-            else:
-                return None
+        from shared.utils import get_entropy, get_majority_vote
```
jzhang-gp, line 338: Can we move the import to the top? Or is there a specific reason we want to import it for every call?

edu-gp: Done

PR #117, opened 2020-07-09
Model details

jzhang-gp:
Models page 
image

Model details page
image

Configs are only shown for models that are ready:
image

The owner of the model is not implemented yet. This means we need to create a new column in the model table but is it the right place to put a user column in a model? What does it mean by having a user owning a model? Should we create another `ModelConfig` table for things like this aside from the deployment config?


edu-gp:
We can add "Owner" in a future PR. A few questions & comments:
1. Why is the "number of useful datapoints" highlighted red there?
2. Can you highlight the "Deployed?" column plz
3. For Pr, Rc, F1, just show the positive class (the second element in the pair). Also, Pr, Rc, F1 changes as you change threshold - that's something we'd want to address in a future PR as well.

jzhang-gp:
Updated UI
image

As per Abe's request:
- Highlight only the font instead of the whole cell
- Highlight ROC_AUC, Precision, Recall and F-score. For the last 3 values, only highlight the positive class.
- Owner dropdown menu to select an owner from existing user.

backend/models.py
```diff
@@ -1,6 +1,10 @@
 import logging
+from collections import namedtuple
+from dataclasses import dataclass
```
edu-gp, line 3: Nice, didn't know about `dataclasses`

backend/models.py
```diff
@@ -59,6 +82,7 @@ def export_new_raw_data():
 
 @bp.route('update_model_deployment_config', methods=['POST'])
 def update_model_deployment_config():
+    logging.error(request.form)
```
edu-gp, line 85: Is this a debug message?

backend/models.py
```diff
@@ -97,3 +121,116 @@ def update_model_deployment_config():
...
+    for model_id, is_approved, is_selected_for_deployment, threshold in \
+            res:
+        deployment_configs_per_model[model_id] = {
+            "is_approved": is_approved,
+            "is_selected_for_deployment": is_selected_for_deployment,
+            "threshold": threshold
+        }
+
+    logging.error(models_per_label)
+    logging.error(deployment_configs_per_model)
```
edu-gp, line 158: Are these debug messages?

db/model.py
```diff
@@ -348,6 +348,8 @@ class Model(Base):
     # Optionally associated with a Label
     label = Column(String, index=True, nullable=True)
 
+    entity_type = Column(String, index=True, nullable=True)
```
edu-gp, line 351: Can you add a default value for `entity_type`? Otherwise I think it'll break other pieces of the code (e.g. when creating a new Model)

backend/models.py
jzhang-gp: I'll remove it.

backend/models.py
jzhang-gp: I'll remove it.

db/model.py
jzhang-gp: This is only temporary. As tested locally, we need this change plus the backfill script to handle the existing models. Once that's done we can update this to be `nullable=False`. 

If we want to add a default type, what would it be? Company?

PR #118, opened 2020-07-13
Setting up the PubSub to publish the message for inference

jzhang-gp:
PubSub for publising the inference output message. 

Tested locally by directly calling the `submit_gcp_inference_on_new_file` method. Triggered a pubsub topic on GCP with a CloudRun function that prints out the message in the log.


backend/api.py
```diff
@@ -38,8 +38,8 @@ def trigger_inference():
     """Trigger inference on a dataset."""
     dataset_name = None
 
-    if 'dataset_name' in request.form:
-        dataset_name = request.form.get('dataset_name')
+    if 'dataset_name' in request.args:
+        dataset_name = request.args.get('dataset_name')
```
edu-gp, line 42: Why make this change? `args` reads from the query string, `form` reads the form body.

train/gs_utils.py
```diff
@@ -104,7 +108,8 @@ def create_deployed_inference(metadata: DeployedInferenceMetadata) -> None:
         gs_copy_file(csv_fname, prod_inference_url, no_clobber=False)
         gs_copy_file(metadata_fname, prod_metadata_url, no_clobber=False)
 
-    publish_message()
+    message = __construct_message(prod_inference_url, prod_metadata_url, ts)
+    publish_message(topic=__prepare_pubsub_topic_name(), message=message)
```
edu-gp, line 112: Can we make this a single call to a function, to hide all the details of how to construct and send a pubsub message?

train/gs_utils.py
```diff
@@ -132,6 +137,25 @@ def build_prod_inference_dataframe(pred_fname, raw_fname, threshold):
...
+            'metadata': prod_meta_url
+        }
+    }
+    return json.dumps(message)
+
+
+def __prepare_pubsub_topic_name():
+    project = os.getenv("GCP_PROJECT_ID")
+    topic = os.getenv("INFERENCE_OUTPUT_PUBSUB_TOPIC")
+    return f'projects/{project}/topics/{topic}'
```
edu-gp, line 161: Can we refactor this to something like "GCPPubSubService"? At least so it can be stubbed and tested

backend/api.py
jzhang-gp: Yeah you are right. Got away when playing with the Postman tool. Reverted.

train/gs_utils.py
jzhang-gp: See the new change.

train/gs_utils.py
jzhang-gp: See the new change.

jzhang-gp:
Created a new module `external_service.py`. We should start to group external calls to modules like this.

backend/external_services.py
```diff
@@ -0,0 +1,44 @@
...
+                'metadata': prod_meta_url
+            }
+        }
+        return json.dumps(message)
+
+    @classmethod
+    def _prepare_pubsub_topic_name(cls):
+        project = os.getenv("GCP_PROJECT_ID")
+        topic = os.getenv("INFERENCE_OUTPUT_PUBSUB_TOPIC")
+        return f'projects/{project}/topics/{topic}'
```
edu-gp, line 44: `GCPPubSubService` is a pretty generic name, but this class is hardcoded with things specific for `inference output`, maybe rename this class? or subclass?

backend/external_services.py
```diff
@@ -0,0 +1,44 @@
...
+                          f"message {message} with exception: {e}")
+            raise e
+
+    @classmethod
+    def _construct_message(cls, prod_inference_url, prod_meta_url, timestamp):
+        message = {
+            'timestamp': timestamp,
+            'environment': 'prod',
+            # ideally we should have multiple stages and not harding this here.
+            'dataset': 'taxonomy_b2c',
```
edu-gp, line 32: Is the dataset meant to be hardcoded?

jzhang-gp: Refactored to `'taxonomy_' + metadata.label`.

backend/external_services.py
jzhang-gp: Refactor to a different version. The GCPPubSubService now takes in the `project_id`, `topic_name` and a function to construct the message plus its **kwargs. The caller is responsible for providing the instruction how a message is constructed.

jzhang-gp:
Some more changes on handling file path. The inference API expects a filename but the file path published by GDP will have path like `pitchbook/2020/07/20/company.jsonl`. We could extract the company.jsonl part but it's better we save the file on our side in similar fashion as `alchemy/data/pitchbook/2020/07/20/company.jsonl`. The API call will also takes in `pitchbook/2020/07/20/company.jsonl` as input. This leads to some changes on how we handle local file path.

train/gcp_job.py
```diff
@@ -222,7 +225,7 @@ def submit_job(model_defns: List[ModelDefn],
 
     # Make sure the filenames are _names_, not _paths_
     files_for_inference = files_for_inference or []
-    files_for_inference = [get_name(f) for f in files_for_inference]
+    # files_for_inference = [get_name(f) for f in files_for_inference]
```
edu-gp, line 228: Just delete this line?

train/gs_url.py
```diff
@@ -1,26 +1,26 @@
...
-    dataset_name_stem = Path(dataset_name).stem
+    dataset_name_stem = retrieve_dataset_name_stem(dataset_name)
     bucket = os.environ.get('GOOGLE_AI_PLATFORM_BUCKET')
     assert bucket, "GCS bucket not defined"
     return f"gs://{bucket}/tasks/{model_uuid}/models/{model_version}/inference/{dataset_name_stem}.pred.npy"
 
 
 def _build_prod_dir(model_uuid, model_version, dataset_name, ts) -> str:
-    stem = Path(dataset_name).stem
+    dataset_name_stem = retrieve_dataset_name_stem(dataset_name)
```
edu-gp, line 20: Can you also check other places we're using the `Path` library? Maybe should replace those places too

train/gcp_job.py
jzhang-gp: Removed.

cloud_function/inference/inference_api_invoker.py
```diff
@@ -0,0 +1,49 @@
...
+
+        storage_client = storage.Client()
+
+        source_bucket_name = path.split("/")[0]
+        source_blob_name = path[len(source_bucket_name) + 1:]
+
+        source_bucket = storage_client.bucket(source_bucket_name)
+        source_blob = source_bucket.blob(source_blob_name)
+
+        destination_bucket_name = 'alchemy-gp'
```
edu-gp, line 29: Can we use the env var `GOOGLE_AI_PLATFORM_BUCKET`

train/train_celery.py
```diff
@@ -135,7 +135,7 @@ def submit_gcp_inference_on_new_file(dataset_name):
 
 
 def submit_gcp_job(model: Model, files_for_inference: List[str],
-                   deploy_metadata: Optional[DeployedInferenceMetadata]):
+                   deploy_metadata: Optional[DeployedInferenceMetadata] = None):
```
edu-gp, line 138: I'm surprised this shows up here, cuz it's already in master: https://github.com/edu-gp/annotation_tool/blob/master/train/train_celery.py#L138

train/train_celery.py
```diff
@@ -177,10 +177,6 @@ def ensure_file_exists_locally(dataset_name: str) -> None:
         local_fname = get_local_data_file_path(dataset_name)
         gs_copy_file(remote_fname, local_fname)
 
-    if dataset_name not in get_all_data_files():
-        raise Exception(
-            f"Dataset {dataset_name} either does not exist or is invalid")
```
edu-gp, line 182: Why remove this?

train/gs_utils.py
```diff
@@ -104,7 +108,30 @@ def create_deployed_inference(metadata: DeployedInferenceMetadata) -> None:
         gs_copy_file(csv_fname, prod_inference_url, no_clobber=False)
         gs_copy_file(metadata_fname, prod_metadata_url, no_clobber=False)
 
-    publish_message()
+    GCPPubSubService.publish_message(
+        project_id=os.getenv("GCP_PROJECT_ID"),
+        topic_name=os.getenv("INFERENCE_OUTPUT_PUBSUB_TOPIC"),
+        message_constructor=_message_constructor_alchemy_to_gdp,
+        dataset='taxonomy_' + metadata.label,
```
edu-gp, line 115: I think the `dataset` should be the `dataset_name` the inference was run on, this isn't the case, right?

train/train_celery.py
jzhang-gp: Hmm the master branch on my local is not up to date. I'll do a pull.

train/train_celery.py
jzhang-gp: Is it not handled by the code block above it? That it downloads the file from remote to local?

cloud_function/inference/inference_api_invoker.py
jzhang-gp: Eh not really. The cloud function code is technically not part of Alchemy repo. Right now I have to manually copy and paste the cloud function code to the Cloud function console for any update. If we have a way to do CICD to both the compute engine and Cloud function, then we may replace it with env var.

train/gs_utils.py
jzhang-gp: Updated to `metadata.dataset_name`, which is assigned to the local variable `dname`

train/train_celery.py
edu-gp: It's a recheck but also, `get_all_data_files` is a special function that also looks at the content of a file to check that it's a valid data file for us to consume. So, in theory, the above code could've downloaded a _patterns_ file instead of _data_ file, and this would've have caught it (hence the "is invalid" part in the exception). I probably should've added a test to guard this ;)

cloud_function/inference/inference_api_invoker.py
edu-gp: Got it...

train/train_celery.py
jzhang-gp: Added back

PR #119, opened 2020-07-14
Need to extract the filename from the absolute path. 

jzhang-gp:
We need absolute path for preparing a model. This is to extract the filename from the path since the `ensure_file_exists_locally` expects just the filename. 

I also need to add `None` to the `submit_gcp_training` method, otherwise it throws exceptions on missing positional arguments.

### DIFF
--- a/train/train_celery.py
+++ b/train/train_celery.py
@@ -1,3 +1,4 @@
+import logging
 import time
 from typing import List, Optional
 from celery import Celery
@@ -61,6 +62,7 @@ def inference(model_dir, raw_file_path):
 
 @app.task
 def submit_gcp_training(label, raw_file_path, entity_type):
+    logging.info("Raw file for the training is " + raw_file_path)
     db = Database.from_config(DevelopmentConfig)
     try:
         model = prepare_next_model_for_label(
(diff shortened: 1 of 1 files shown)


train/train_celery.py
```diff
@@ -141,7 +143,8 @@ def submit_gcp_job(model: Model, files_for_inference: List[str],
             results to be deployed.
     """
     for dataset_name in files_for_inference:
-        ensure_file_exists_locally(dataset_name)
+        filename = dataset_name.split("/")[-1]
```
edu-gp, line 146: You might as well change `files_for_inference` to just file names instead of paths. Maybe better to rename it to `datasets_for_inference`.

Also, please use something like https://github.com/edu-gp/annotation_tool/blob/master/train/no_deps/paths.py#L70 to get the filename - just so we have a consistent way of doing it!

jzhang-gp: `files_for_inference` are paths because there are other places that require paths. I remember before `submit_gcp_training` is called, in a method to `prepare_model`, we need the path to load jsonl files. So somewhere in between we need to extract the filenames. 

I'll change the way we get the filename to the one you mentioned.

edu-gp: Hm, it shouldn't need the path in theory. I can take a look later.

PR #120, opened 2020-07-15
Remove poll loop

edu-gp:
For checking the status of AI Platform jobs, we remove poll loop, and replace it with a job that's placed back onto the queue 60 seconds later.

This also cancels runaway jobs (jobs that take 1 hour or more).

Finally, this includes a stricter check on AI Platform job states, so the worker would only place itself back onto the queue if the AI Platform job is still in progress.

### DIFF
--- a/train/gcp_celery.py
+++ b/train/gcp_celery.py
@@ -1,4 +1,4 @@
-import time
+import logging
 from typing import Optional
 from celery import Celery
 from train.gcp_job import GoogleAIPlatformJob, download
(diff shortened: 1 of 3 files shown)


train/gcp_celery.py
```diff
@@ -15,33 +15,59 @@
...
+            logging.info(f"Done AI Platform job_id={job_id}")
+            on_job_success(job, metadata_dict)
+        elif poll_count > 60:
+            # A job is taking too long. Try to cancel it.
+            try:
+                logging.info(f"Cancel AI Platform job_id={job_id}")
+                job.cancel()
+            except Exception:
+                # If an error occurred, we can try again later.
+                pass
```
jzhang-gp, line 55: Can we log the exception too?

train/gcp_job.py
```diff
@@ -338,6 +342,18 @@ def get_model_defns(self) -> List[ModelDefn]:
...
+    def cancel(self) -> bool:
+        """Cancel a running job
+        Returns True if the job was cancelled successfully
+        Raises:
+            Exception if job_id is missing
+            Exception if a job has already completed
+        """
+        job_id = self.response.get("jobId")
+        assert job_id, "Cannot cancel job, missing job_id"
+        run_cmd(f'gcloud ai-platform jobs cancel {job_id}')
```
jzhang-gp, line 354: Is there a particular reason we are using the gcloud tool with cmd instead of the python client to call GCP service? My main concern is that error details from these command may not be properly logged or captured.

edu-gp: Good call. I've been using the gcloud tool for everything else - would be good to migrate them all to using the python client. I'll add a task to fix this later.

train/gcp_celery.py
edu-gp: Done

PR #123, opened 2020-07-23
Bugfix for the models page

jzhang-gp:
Bugfix for the models page including the UI style change. We need to use `container-fluid` for the page to show the table properly.

### DIFF
--- a/backend/models.py
+++ b/backend/models.py
@@ -210,10 +210,19 @@ def _collect_model_data_rows():
         if chosen_model and chosen_model.is_ready():
             test_metrics = chosen_model.get_metrics().get('test', {})
 
-            row.roc_auc = test_metrics.get('roc_auc', None)
-            row.pr = test_metrics.get('precision', None)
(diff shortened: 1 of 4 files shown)


backend/models.py
```diff
@@ -175,55 +173,56 @@ def _collect_model_data_rows():
...
                                    'num')). \
             join(ClassificationAnnotation). \
             filter(ClassificationAnnotation.label == label,
                    ClassificationAnnotation.value != AnnotationValue.NOT_ANNOTATED). \
             group_by(User.username).order_by(desc('num')).all()
-        majority_annotator = res[0][0]
+        if annotators_sorted and len(annotators_sorted) > 0:
+            majority_annotator = annotators_sorted[0][0]
+        else:
+            majority_annotator = None
```
edu-gp, line 194: Just a general comment - it probably would be easier to read & easier to test if we refactored this function into a few helper functions like `get_chosen_model` and `get_majority_annotator`

PR #125, opened 2020-07-28
Update the cloud function as we are using HTTP trigger 

jzhang-gp:
Change how we handle the request as we are processing a HTTP request, not a pubsub message.

Tested on GCP and trigger is now working properly.

### DIFF
--- a/cloud_function/inference/inference_api_invoker.py
+++ b/cloud_function/inference/inference_api_invoker.py
@@ -6,16 +6,21 @@ import requests
 from google.cloud import storage, secretmanager
 
 
-def handler(event, context):
+def handler(request):
     """Triggered from a message on a Cloud Pub/Sub topic.
     Args:
          event (dict): Event payload.
          context (google.cloud.functions.Context): Metadata for the event.
     """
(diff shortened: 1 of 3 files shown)


cloud_function/inference/inference_api_invoker.py
```diff
@@ -6,16 +6,21 @@
...
          context (google.cloud.functions.Context): Metadata for the event.
     """
-    pubsub_message = base64.b64decode(event['data']).decode('utf-8')
-    print(pubsub_message)
+    request_json = request.get_json()
+    message = None
+    if request.args and 'message' in request.args:
+        message = request.args.get('message')
+    elif request_json and 'message' in request_json:
+        message = request_json['message']
```
edu-gp, line 20: Do you need request.args and request_json? Would it be easier to maintain if there's only 1 format?

jzhang-gp: This is generated by the Cloud Function http handler template. I'm assuming there's a reason behind this. I'd rather keep it for now.

PR #127, opened 2020-07-28
Golden annoations

jzhang-gp:
Golden annotation changes. 

The current majority votes use two different logic but it's not an easy thing to consolidate into one. The scenario is different. One involves query but the other is deeply in the pandas data frame. 

The changes here keep the old logic but augment them with the weight in the annotations.

I have not removed the old methods yet. Once we are OK with the new ones, I'll deprecate them in subsequent commits.


backend/annotations_utils.py
```diff
@@ -14,6 +15,10 @@ def _parse_list(form: dict, key: str) -> List:
 def parse_form(form: dict):
     user = form.get('user')
     label = form.get('label')
+
+    golden_annotations = form.get("golden_annotations", None)
+    golden_annotations = True if golden_annotations else False
```
edu-gp, line 20: Rename to `is_golden_annotations` or `is_golden` please

backend/tasks.py
```diff
@@ -307,45 +307,52 @@ def download_prediction():
...
+            _df = _df.rename(columns={'value': username,
+                                      'entity': 'domain',
+                                      'weight': username+"_vote_weight"})
             # Merge it with the main dataframe.
             df = df.merge(_df, on='domain', how='left')
 
         # Compute some statistics of the annotations
         # Only consider the columns with the user annotations
-        df_annos = df.iloc[:, n_cols:]
+        # df_annos = df.iloc[:, n_cols:]
```
edu-gp, line 352: Delete this line

backend/templates/annotations/bulk.html
```diff
@@ -42,6 +42,11 @@ <h1>Bulk Upload Annotations</h1>
                 </select>
             </div>
 
+            <div class="form-group">
+                <label for="label">Set as GOLDEN ANNOTATIONS</label><br/>
```
edu-gp, line 46: Add a bit of help text to tell the user what golden annotations are

shared/utils.py
```diff
@@ -176,3 +178,15 @@ def get_majority_vote(annos: List[Optional[int]]):
...
+def get_majority_vote_v2(annos: List[Optional[int]]):
+    valid_votes = [x for x in annos if x != 0 and not pd.isna(x)]
+    if len(valid_votes) > 0:
+        total = sum(valid_votes)
+        if total != 0:
+            return 1 if total > 0 else -1
+        else:
+            return random.choice([-1, 1])
+    else:
+        return None
```
edu-gp, line 192: Please remove all the `_v2` and just replace this with the default implementation. (It also makes it easier to review the differences)

backend/annotations_utils.py
jzhang-gp: Done.

backend/tasks.py
jzhang-gp: Done.

backend/templates/annotations/bulk.html
jzhang-gp: Done.

shared/utils.py
jzhang-gp: Done.

jzhang-gp:
Added a fix for the bug we had. In the old query we did not guard the cases when the count is equal for the positive and negative case. In this case we want to keep just one so the change includes a distinct keyword in the query.

backend/annotations.py
```diff
@@ -1,3 +1,5 @@
+import logging
```
edu-gp, line 1: Is this import needed?

backend/annotations_utils.py
```diff
@@ -1,3 +1,4 @@
+import logging
```
edu-gp, line 1: Is this import needed?

backend/tasks.py
```diff
@@ -312,40 +311,53 @@ def download_prediction():
         q = db.session.query(
             User.username,
             ClassificationAnnotation.entity,
-            ClassificationAnnotation.value
+            ClassificationAnnotation.value,
+            ClassificationAnnotation.weight,
+            ClassificationAnnotation.value * ClassificationAnnotation.weight
```
edu-gp, line 316: What's the logic behind this `weighted_value` calculation? What happens if the value is `-2`? Why not just use `weight` alone?

backend/templates/annotations/bulk.html
```diff
@@ -42,6 +42,11 @@ <h1>Bulk Upload Annotations</h1>
                 </select>
             </div>
 
+            <div class="form-group">
+                <label for="label">Set as GOLDEN ANNOTATIONS (The ground truth that overrides everyone's annotations).</label><br/>
+                <input type="checkbox" name="golden_annotations" value="true" />
```
edu-gp, line 47: Can we call this `is_golden` as well? Just to be consistent.

db/model.py
```diff
@@ -221,27 +222,30 @@ def majority_vote_annotations_query(dbsession, label):
...
     ).group_by(q1.c.entity)
 
     q2 = q2.cte('max_query')
 
-    query = dbsession.query(
-        q1.c.entity,
+    # in case the weights are for the positive and negative classess, we need
+    # select only one
+    query = dbsession.query(distinct(
+        q1.c.entity),
```
edu-gp, line 245: distinct(entity) should be on one line, so it's easier to read.

db/model.py
```diff
@@ -221,27 +222,30 @@ def majority_vote_annotations_query(dbsession, label):
...
         q1.c.value,
-        q1.c.count
-    ).join(q2, (q1.c.entity == q2.c.entity) & (q1.c.count == q2.c.count))
+        q1.c.weight
+    ).join(q2, (q1.c.entity == q2.c.entity) & (q1.c.value == q2.c.value))
```
edu-gp, line 248: Do you mean `q1.c.weight == q2.c.weight`?

Also, I wonder if `weight` being a float can cause any numerical issues in this direct `==` comparison?

shared/utils.py
```diff
@@ -165,16 +167,19 @@ def get_entropy(annos: List[Optional[int]], eps=0.0001):
...
-
-    if len(cnt):
-        return cnt.most_common(1)[0][0]
+    valid_votes = [x for x in annos if x != 0 and not pd.isna(x)]
+    if len(valid_votes) > 0:
+        total = sum(valid_votes)
+        if total != 0:
+            return 1 if total > 0 else -1
+        else:
+            return random.choice([-1, 1])
```
edu-gp, line 183: Looking at the tests, I understand what you're doing here now. Can we instead pass in a list of tuples of (label, weight), rather than a list of ints of `value * weight`? If you're worried about backwards compatibility, you can also create a new function `get_weighted_majority_vote`. This code now is tightly coupled with the definition of a positive (+1) and negative (-1) label, which is not desirable for a utility function.

backend/annotations_utils.py
jzhang-gp: Removed.

backend/annotations.py
jzhang-gp: Removed.

backend/templates/annotations/bulk.html
jzhang-gp: Done.

db/model.py
jzhang-gp: Done.

shared/utils.py
jzhang-gp: Done.

backend/tasks.py
jzhang-gp: Done.

db/model.py
jzhang-gp: Is there anything missing by only checking the value? Why compare the weight? 

Weight can be equal among different label values and cause the output file to have duplicate entries. 

If we have to use weight, we need to edit the query and limit to return only 1 row. 

I have added some comments about what each cte could return. Let me know if you find a bug.

edu-gp: I read it more carefully and I think the confusion comes from the fact the queries I wrote before is different from the queries you wrote here. To explain what I did before:

- `q1` finds the count per `(entity, value)`
- `q2` finds the max count for each `entity`.
- `query` retains only rows of `q1` where it matches `entity, max(count)` from `q2`.

Looking at your change, you effectively changed `count` to `weight`... but your `q2` now contains `value`, I'm actually not sure what that would return. Is that the `value` associated with the max weight? If that's the case you don't even need `query`.

jzhang-gp: Updated a new query. This one uses row_number to avoid the duplicate entities in the result set.

PR #130, opened 2020-07-29
Use post request json param. 

jzhang-gp:
I have to change the function to use the `json` param in the `requests.post` method call. Somehow the form data does not show up (payload is empty) when I use the `data` param and it only happens with the cloud function in production. Locally the request works just fine. If you have some idea on why this is happening, please let me know.

### DIFF
--- a/backend/api.py
+++ b/backend/api.py
@@ -47,13 +47,18 @@ def trigger_inference():
     """Trigger inference on a dataset."""
     dataset_name = None
 
-    if 'dataset_name' in request.form:
-        dataset_name = request.form.get('dataset_name')
+    json_data = request.get_json()
+
+    if json_data:
+        request_id = json_data.get('request_id')
+        logging.info("Handling request " + request_id)
+        dataset_name = json_data.get('dataset_name', None)
 
     if dataset_name:
         run_inference_on_data(dataset_name)
         return "OK", 200
     else:
+        logging.error(f"[{request_id}] Missing dataset name. Abort...")
         return abort(400)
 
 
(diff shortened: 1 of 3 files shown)


backend/api.py
```diff
@@ -47,13 +47,18 @@ def trigger_inference():
...
     dataset_name = None
 
-    if 'dataset_name' in request.form:
-        dataset_name = request.form.get('dataset_name')
+    json_data = request.get_json()
+
+    if json_data:
+        request_id = json_data.get('request_id')
+        logging.info("Handling request " + request_id)
+        dataset_name = json_data.get('dataset_name', None)
```
edu-gp, line 55: Do all the tests still pass when you switched it from form to json payload?

cloud_function/inference/inference_api_invoker.py
```diff
@@ -1,16 +1,20 @@
...
-    """Triggered from a message on a Cloud Pub/Sub topic.
+    """Responds to any HTTP request.
     Args:
-         event (dict): Event payload.
-         context (google.cloud.functions.Context): Metadata for the event.
+        request (flask.Request): HTTP request object.
+    Returns:
+        The response text or any set of values that can be turned into a
+        Response object using
+        `make_response <http://flask.pocoo.org/docs/1.0/api/#flask.Flask.make_response>`.
```
edu-gp, line 17: This looks like more generic comments than before. Was that intentional?

jzhang-gp: Remove.

PR #131, opened 2020-08-01
Rework metrics computation and UI, and show a confusion matrix table

edu-gp:
Screen Shot 2020-08-01 at 2 35 23 PM

The UI for model metrics have been touched up. Underneath, we're also computing the metrics from scratch (instead of relying on the model metrics computed when it was trained). This allows us to, in the near future, recompute the metrics when the user changes the threshold.

Note, one caveat when computing the metrics from scratch is that, when there has been an issue (e.g. in training the model, mistake in moving data) we're not able to load the inference for some data points. So I show the warning, as in the above picture, to help us catch those issues and debug them.

The bulk of the logical changes are in `train/no_deps/metrics.py`. I also refactored a `_prepare_data` function out of `train/no_deps/run.py`for reuse. The rest are pretty much UI template changes and tests.

Still left to do, in future PRs:
1. Address the efficiency issue (probably using cache). This significantly slows down page loads to a few seconds.
2. Reuse this UI in the model page.
3. There's some clean-up refactoring work left to do (see the `TODO` in `Model.compute_metrics`)


db/model.py
```diff
@@ -460,9 +455,28 @@ def get_len_data(self):
...
+        """See train.no_deps.compute_metrics"""
+
+        # TODO this code is temporarily here; will refactor it later when I
+        # refactor the code for export_inference as well.
+        import pandas as pd
+        df = []
+        for fname in self.get_inference_fnames():
+            df.append(self.export_inference(fname, include_text=True))
+
+        if df:
```
jzhang-gp, line 470: So empty list evaluate to `True`??

db/model.py
```diff
@@ -460,9 +455,28 @@ def get_len_data(self):
...
+            df = pd.concat(df, axis=0)
+        else:
+            # Make a dummy dataframe, in case no inference data is found.
+            df = pd.DataFrame(columns=['text', 'probs'])
+
+        version_dir = self.dir(abs=True)
+
+        from train.no_deps.metrics import compute_metrics as _compute_metrics
```
jzhang-gp, line 478: Can we move this to the top of the file in the import section?

train/no_deps/metrics.py
```diff
@@ -0,0 +1,129 @@
...
+    config_fname = _get_config_fname(version_dir)
+    data_fname = _get_exported_data_fname(version_dir)
+
+    _, _, X_train, y_train, X_test, y_test = \
+        _prepare_data(config_fname, data_fname)
+
+    not_found = []
+
+    def _get_pred_result(X, y, df):
+        nonlocal not_found
```
jzhang-gp, line 72: Is it possible to not use side-effect code and "global variable" like this?

train/no_deps/metrics.py
```diff
@@ -0,0 +1,129 @@
...
+        res = pd.DataFrame(zip(X, y), columns=['text', 'y'])
+        res = res.merge(df, on='text', how='left')
+
+        not_found += list(res[res['probs'].isna()]['text'])
+
+        res = res.dropna(subset=['probs'])
+
+        return res
+
+    def _get_stats(res, threshold):
```
jzhang-gp, line 83: Can we also extract these inner functions to their own places? The `compute_metrics` function is getting too big.

db/model.py
edu-gp: empty list evals to `False`, in that case we make a dummy dataframe, as per the `else` case

train/no_deps/metrics.py
edu-gp: Done

db/model.py
edu-gp: Done

train/no_deps/metrics.py
edu-gp: Done

PR #132, opened 2020-08-05
Updating the dataset name in the message

jzhang-gp:
There is a gap on our understanding of the `dataset` attribute in the PubSub message published. We can chat offline on the details. 

The other team are using the dataset attribute more like data-category to shard the entries. For example, these data are from `alchemy` and these are from `SF` etc. It's not specific filename they are targeting at and it's captured in the prediction filename anyway.


train/gs_utils.py
```diff
@@ -112,7 +112,7 @@ def create_deployed_inference(metadata: DeployedInferenceMetadata) -> None:
         project_id=os.getenv("GCP_PROJECT_ID"),
         topic_name=os.getenv("INFERENCE_OUTPUT_PUBSUB_TOPIC"),
         message_constructor=_message_constructor_alchemy_to_gdp,
-        dataset=dname,
+        dataset='alchemy-inference',
```
edu-gp, line 115: 1. If we must use this, can we make it an env var instead of hard-coding it here?
2. Can you also make a comment here about why `dataset` is not `dname`, as you have written in the PR?
3. The reason to keep track of `dname` in the published message is so 1) external systems know where to look for the specific files for this inference run and 2) in the future, if we want to debug something, we can trace back to where the data came from. Are these goals still met?

jzhang-gp: One more to do here: Need to update the metadata.json field to 
```
{
"timestamp": 1234, 
"model_uuid": "gsdfdsfsg1233", 
"model_version": 2, 
"label": "TestingV2", 
"threshold": 0.6, 
"dataset_name": "XXX-inference", 
"filename": "XXXXX-2020-08-05-00-00.jsonl"
}
```

jzhang-gp: >     1. If we must use this, can we make it an env var instead of hard-coding it here?
> 
>     2. Can you also make a comment here about why `dataset` is not `dname`, as you have written in the PR?
> 
>     3. The reason to keep track of `dname` in the published message is so 1) external systems know where to look for the specific files for this inference run and 2) in the future, if we want to debug something, we can trace back to where the data came from. Are these goals still met?

1. Are we going to have a different value other than 'alchemy-inference'? Is a constant variable enough? I'm OK with an env variable too but what's the usecase?
2. Added.
3. The name of the file is part of the inference file url as `gs://{BUCKET_NAME}/prod/14d1f873dcXXXXXXa3773253/2/{THE_FILENAME}-{DATE}/{TIME_STAMP}/inference.csv"`. So yes we know where the data come from. I will also add the filename in the metadata json file so the other team also knows the source.

edu-gp: 1. To keep it in the spirit of a project other orgs can take and use :) (We're steering further from that but I'm still aspiring to this goal, if possible!)
2. 👍 
3. Okay, sounds good. As long as we have a way to trace and debug, when things (inevitably) go wrong.

jzhang-gp: Done.

PR #133, opened 2020-08-06
Bulk upload positive

jzhang-gp:
Bulk upload page for companies and labels with the same annotation value.

### DIFF
--- a/backend/annotations.py
+++ b/backend/annotations.py
@@ -1,11 +1,13 @@
+import logging
+
 from flask import (
     Blueprint, flash, redirect, render_template, request
 )
 from db.model import (
     db, EntityTypeEnum,
-    ClassificationAnnotation, User, get_or_create
-)
-from .annotations_utils import parse_form
+    ClassificationAnnotation, User, get_or_create,
+    AnnotationValue)
+from .annotations_utils import parse_form, parse_bulk_upload_v2_form
 
 from .auth import auth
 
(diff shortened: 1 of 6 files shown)


backend/annotations.py
```diff
@@ -26,6 +28,69 @@ def index():
     return render_template('annotations/index.html')
 
 
+@bp.route('/bulk_v2', methods=['GET'])
+def bulk_v2():
```
edu-gp, line 32: Can we use a more descriptive name than `bulk_v2`?

backend/annotations.py
```diff
@@ -104,11 +145,52 @@ def bulk_post():
...
 
     except Exception as e:
+        logging.error(e)
         flash(str(e))
         return render_template('annotations/bulk.html')
     else:
         return redirect(redirect_to)
+
+
+def __upsert_annotations(entity_type, entity, label, user, value):
```
edu-gp, line 159: 1. Please add test to `__upsert_annotations`
2. Would it be ok to use 1 underscore? e.g. `_upsert_annotations`

backend/annotations.py
jzhang-gp: Done. Changed to `bulk_upload_positive_annotations`.

backend/annotations.py
jzhang-gp: Done.

PR #136, opened 2020-09-01
Cache metrics computation

edu-gp:
This is to fix a previous issue where computing the metrics of a model is slow, making the UI not useable.

Computing metrics is slow because it has to load in a bunch of files in order to find the predicted values of the train & test sets. (This could've been a lot simpler & faster if we had a different data storage design - a discussion we should have later)

After much deliberation, I've decided the best approach is to do the metrics computation on the first page load, cache the result on disk, so subsequent page loads are instant. This PR also includes some left-over refactoring.

### DIFF
--- a/db/model.py
+++ b/db/model.py
@@ -1,8 +1,9 @@
+from typing import Optional, List
 import logging
 import copy
 import os
 import urllib.parse
-from typing import List
+import pandas as pd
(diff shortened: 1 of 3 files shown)


edu-gp:
Merging to unblock me

PR #137, opened 2020-09-03
(WIP) Deploy with docker-compose

edu-gp:
Switch over deployment from supervisord to docker-compose. At a later time, we can deprecate supervisord completely.

To set up a new deployment:

1. Create folder `__filestore`, and `/models, /raw_data, /training_data` under it
2. Adjust `.env` and `docker-compose` file (note all the env vars)
3. Build the training Docker image and point to it in the `.env` file
4. Build the base Docker image `docker build -t alchemy-base .`, this one has the common deps
5. Run `docker-compose up --build` (it will build each image and run everything) go into the `backend` containers and run `alembic upgrade head` to migrate the database.

```
docker build -t alchemy-base .
docker-compose up --build -d
```

```
docker exec -it <container_id> bash
alembic upgrade head
```

In this setup, it uses a local postgres database. In prod, we can adjust the env vars to make it point to the prod db.

On every code change
```
docker-compose down
docker-compose up --build -d
```

To monitor, you can use docker-compose to view logs and stats.

---

Note new addition env vars
```
FILESTORE_LOCAL_DIR=...
GCLOUD_LOCAL_DIR=...
GOOGLE_APPLICATION_CREDENTIALS_LOCAL_DIR=...

REDIS_HOST=redis
POSTGRES_DB=alchemy
POSTGRES_USER=username
POSTGRES_PASSWORD=password
```

and these two old env vars have to correspond
```
ALCHEMY_DATABASE_URI=postgres://username:password@db:5432/alchemy
DB_URL_FOR_MIGRATION=postgres://username:password@db:5432/alchemy
```

---

Also included here are some changes to deprecate old code...


.env.example
```diff
@@ -1,27 +1,51 @@
-ANNOTATION_TOOL_BACKEND_PASSWORD=secret
-ANNOTATION_TOOL_FRONTEND_SECRET=fegGA9r1f13f9jf3
+# Basics
+ALCHEMY_ENV=dev
 ANNOTATION_TOOL_FRONTEND_SERVER=http://localhost:5001
-ANNOTATION_TOOL_MAX_PER_ANNOTATOR=100
-ANNOTATION_TOOL_MAX_PER_DP=3
+ANNOTATION_TOOL_FRONTEND_SECRET=fegGA9r1f13f9jf3
+ANNOTATION_TOOL_BACKEND_PASSWORD=secret
```
jzhang-gp, line 5: Are these example passwords?

.env.example
```diff
@@ -1,27 +1,51 @@
...
+API_TOKEN=foobar1234
+FLOWER_BASIC_AUTH=foo:bar
```
jzhang-gp, line 7: Same here.

docker-compose.yml
```diff
@@ -0,0 +1,90 @@
+version: '3.4'
```
jzhang-gp, line 1: I may need a walk through on this file. It looks like what the `supervisord` is doing but how does the images work? What are they referring to?

jzhang-gp (changes requested):
Let's chat to walk through how to use this. I'm a bit rusty on docker-composer :)

.env.example
edu-gp: Yes they're not what we actually use

.env.example
edu-gp: Yep just examples

docker-compose.yml
edu-gp: I'll walk you through this file tomorrow. Here's a complete reference: https://docs.docker.com/compose/compose-file/

PR #138, opened 2020-09-16
Optionally enable Cloud Logging via an env var

edu-gp:
Cloud Logging caused some permission issues when deploying at other companies (where they didn't need cloud logging), so adding this env var to disable it by default.


jzhang-gp (approved):
LGTM

PR #139, opened 2020-09-16
Document optional Nginx setup, for when we need HTTPS

edu-gp:
Sometimes companies want to expose Alchemy on the public web via HTTPS. Here is how to set that up, even without your own domain and certificate.


jzhang-gp (approved):
LGTM

PR #140, opened 2020-09-25
Refactor GoogleAIPlatformJob to initialize with a job_id

edu-gp:
The `GoogleAIPlatformJob` used to be initialized with dict. This is quite confusing because you wouldn't know what to pass in, and it's error-prone. We can actually just initialize it with a `job_id` string instead.

Also refactored methods that call into the `gcloud ai-platform` cli into their own functions, so we can at a later time easily replace these with proper API calls.

NEXT: After this PR, there are some more cleanup to do in `train/gcp_job.py`


jzhang-gp (approved):
LGTM

PR #141, opened 2020-09-28
Introduce ModelStorageManager and DatasetStorageManager

edu-gp:
A lot of indirection and duplicate code exist to handle moving _model data_ and _raw datasets_ to and from GCS. In this change, we introduce `ModelStorageManager` and `DatasetStorageManager` under `storage_manager.py` to handle the movement of data. Each StorageManager has conceptually a "local" directory (either on the server, or on the AI Platform), and a "remote" directory (on GCS). They expose interfaces like "download" and "upload". Elsewhere -- in submitting a new AI Platform job, and in starting training within an AI Platform job -- we can use these StorageManagers to simplify a lot of the code.

This also introduces some changes that renames `fname` to either `dataset` or `path`, where appropriate. The definition of a `dataset` is the filename with `.jsonl` ending, e.g. `data.jsonl`. A `path` is the file path, e.g. `/home/user/data.jsonl`. In the codebase, `fname` is loosely defined and sometimes can be either a `dataset` or a `path`, and is a source of confusion and unnecessary code. This PR addresses some of the cases where it's relevant, but not all.

TODO:
- [x] Fix tests
- [x] Run a manual test

### DIFF
--- a/tests/unit/test_train_no_deps_paths.py
+++ b/tests/unit/test_train_no_deps_paths.py
@@ -1,10 +1,10 @@
 from train.no_deps.paths import (
-    _inference_fnames_to_original_fnames
+    _inference_fnames_to_datasets
 )
 
 
-def test__inference_fnames_to_original_fnames():
-    fnames = _inference_fnames_to_original_fnames([
+def test__inference_fnames_to_datasets():
+    fnames = _inference_fnames_to_datasets([
         'blah.pred.npy',
         'blah.blah.pred.npy',
         'invalid_file.txt',  # Invalid files are ignored
--- a/tests/unit/test_train_no_deps_paths.py
+++ b/tests/unit/test_train_no_deps_paths.py
@@ -4,14 +4,14 @@ from train.no_deps.paths import (
 
 
 def test__inference_fnames_to_datasets():
-    fnames = _inference_fnames_to_datasets([
+    datasets = _inference_fnames_to_datasets([
         'blah.pred.npy',
         'blah.blah.pred.npy',
         'invalid_file.txt',  # Invalid files are ignored
         '/home/dir/ok.pred.npy']
     )
 
-    assert fnames == [
+    assert datasets == [
         'blah.jsonl',
         'blah.blah.jsonl',
         'ok.jsonl'
--- a/train/no_deps/gcp_run.py
+++ b/train/no_deps/gcp_run.py
@@ -1,52 +1,10 @@
+# This is the entrypoint to an AI Platform job.
+
 import os
 import tempfile
 from pathlib import Path
-from .run import (
-    train_model as _train_model,
-    inference as _inference,
-    build_inference_cache
-)
-from .utils import (
-    gs_copy_dir as copy_dir,
-    gs_copy_file as copy_file
-)
-from .paths import (
-    _get_all_inference_fnames,
-    _inference_fnames_to_original_fnames
-)
-
-
-def train_model(local_dir, remote_dir, force_retrain=False):
-    # Download config and data
-    copy_dir(remote_dir, local_dir)
-
-    # Train Model
-    _train_model(local_dir, force_retrain=force_retrain)
-
-    # Upload trained model
-    copy_dir(local_dir, remote_dir)
-
-
-def download_files_to_local(remote_data_dir, fnames, local_data_dir):
-    remote_fnames = [f'{remote_data_dir}/{fname}' for fname in fnames]
-
-    local_fnames = []
-    for fname in remote_fnames:
-        local_fname = os.path.join(local_data_dir, Path(fname).name)
-        copy_file(fname, local_fname)
-        local_fnames.append(local_fname)
-    return local_fnames
-
-
-def inference(local_dir, remote_dir, local_fname, inference_cache=None):
-    # Download a trained model
-    copy_dir(remote_dir, local_dir)
-
-    # Run Inference - results saved in local_dir
-    _inference(local_dir, local_fname, inference_cache=inference_cache)
-
-    # Upload results
-    copy_dir(local_dir, remote_dir)
+from .run import train_model, inference, build_inference_cache, InferenceCache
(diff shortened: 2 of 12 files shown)


train/no_deps/storage_manager.py
```diff
@@ -0,0 +1,71 @@
...
+    remote counterpart on GCS."""
+
+    def __init__(self, remote_model_dir: str, local_model_dir: str):
+        """
+        Inputs:
+            gcs_bucket: The bucket (or prefix) on GCS to store model files and
+                datasets. e.g. "alchemy". Note it does not include "gs://".
+        """
+        self.remote_dir = remote_model_dir
+        self.local_dir = local_model_dir
```
jzhang-gp, line 18: The comments here don't seem to match with the arguments.

train/no_deps/gcp_run.py
```diff
@@ -68,52 +26,56 @@ def run(remote_model_dirs, remote_data_dir, infer_fnames, force_retrain, eval_ba
...
+                    # Cache is not nessesary but makes inference on incremental
+                    # data updates a lot faster.
+                    if inference_cache is None:
+                        inference_cache = \
+                            build_inference_cache(msm.local_dir, dsm)
+
+                    # Run Inference - results saved in local_model_dir
+                    # Note the inference_cache is updated for each new file.
+                    inference(msm.local_dir, dataset_local_path,
+                              inference_cache=inference_cache)
```
jzhang-gp, line 75: We are building a cache per inference file. Would it make sense to build a cache for all inference files first?

train/no_deps/storage_manager.py
```diff
@@ -0,0 +1,71 @@
+import os
+from pathlib import Path
+from .utils import run_cmd
+from train.no_deps.utils import gs_copy_file
+
+
+class ModelStorageManager:
+    """A logical interface to a model's assets. A ModelStorageManager instance
+    provides methods to work with a local version of the model's assets and its
+    remote counterpart on GCS."""
```
jzhang-gp, line 9: Can we also add some contexts on what's in the model's asset?

train/no_deps/run.py
```diff
@@ -126,78 +128,100 @@ def plot_results(version_dir, fname, inference_results) -> None:
...
             # By now, all elements in text should have some results,
             # so we can safely use [ ] to access.
-            raw.append(inference_cache[t])
+            raw.append(inference_cache.get(t))
     else:
         # If not using the cache, we just predict on all text.
         _, raw = model.predict(text)
-        to_infer = text
 
     inference_results = InferenceResults(raw)
```
jzhang-gp, line 217: Is there a python file for the `InferenceResults` class?

jzhang-gp (changes requested):

tests/unit/test_train_no_deps_paths.py
```diff
@@ -1,10 +1,10 @@
 from train.no_deps.paths import (
-    _inference_fnames_to_original_fnames
+    _inference_fnames_to_datasets
 )
 
 
-def test__inference_fnames_to_original_fnames():
-    fnames = _inference_fnames_to_original_fnames([
+def test__inference_fnames_to_datasets():
+    fnames = _inference_fnames_to_datasets([
```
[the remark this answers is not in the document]
edu-gp: This function converts an input like `/home/dir/ok.pred.npy` or `ok.pred.npy' to the corresponding dataset `ok.jsonl`. What do you think is a good name for it?

train/no_deps/storage_manager.py
edu-gp: Fixed

train/no_deps/gcp_run.py
edu-gp: It does build inference for all files it has already ran inference on. Is that what you meant?

Let's say the model has already ran inference on "a.jsonl" and "b.jsonl" on previous runs, and now we want it to run on "c.jsonl" and "d.jsonl". So, the `infer_fnames` would be `["c.jsonl", "d.jsonl"]`.

Upon the first loop, `dataset = "c.jsonl"`, because `inference_cache` is None, we call `build_inference_cache`, which builds a cache with the previously inferenced results "a.jsonl" and "b.jsonl". Then, when we call `inference(...)`, the `inference_cache` now holds the results from "a.jsonl", "b.jsonl", and "c.jsonl".

On the second loop, `dataset = "d.jsonl"`, and we skip to call `inference(...)` right away because an `inference_cache` already exists.

train/no_deps/storage_manager.py
edu-gp: This isn't quite the right place to do it, so I left a note to point it to `paths.py`, where I added more info.

train/no_deps/run.py
edu-gp: https://github.com/georgianpartners/annotation_tool/blob/master/train/no_deps/inference_results.py

train/no_deps/gcp_run.py
jzhang-gp: OK got it. So `a.jsonl` and `b.jsonl` are already inside the `local_dir` when building the cache.

tests/unit/test_train_no_deps_paths.py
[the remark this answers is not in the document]
jzhang-gp: Oh I mean on line 7, from `fnames` to `datasets`.

jzhang-gp (approved):
LGTM

train/no_deps/gcp_run.py
edu-gp: Yep. In the actual function `build_inference_cache`, it would check if `a.jsonl` and `b.jsonl` exists locally, and download them from GCS as needed.

tests/unit/test_train_no_deps_paths.py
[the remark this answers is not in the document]
edu-gp: Ah ok. Fixed.

PR #142, opened 2020-09-30
Remove unused deprecated code

jzhang-gp:
Remove some unused code. There are more but this change has been sitting in my local branch for so long.


edu-gp:
LGTM

PR #143, opened 2020-10-05
Changes made to deploy successfully on local machine

jzhang-gp:
Need to make a few small changes to deploy the code on local machine and connect to the postgres server in docker locally.


dhuang-gp (approved):
To me, `.env.example` looks more like a template. I would provide real example file with things filled out. I think a sample one would be helpful for users, because for example the format of`GOOGLE_AI_PLATFORM_DOCKER_IMAGE_URI` is actually not a URI, so that might confuse users.

PR #144, opened 2020-10-06
Utility changes to make development easier

jzhang-gp:
Some utility and comment changes to allow easier development locally. Basically with this volume attachment, the code changes can take effect immediately without restarting the services.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,7 +12,7 @@ x-volumes:
   # immediately without restarting the docker service. Note that this will
   # restart all services. If you want to narrow down the scope of the change,
   # specify a desired folder like $LOCAL_PROJECT_ROOT/db:/app/db
-  - $LOCAL_PROJECT_ROOT:/app
+  - $PWD:/app
 
 services:
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,9 +4,15 @@ version: '3.4'
 x-volumes:
   &default-volumes
   - $GCLOUD_LOCAL_DIR:/root/.config/gcloud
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -4,9 +4,15 @@ version: '3.4'
...
-  # By default the credential location is under $GCLOUD_LOCAL_DIR/credentials. Unless it is customized, there is no need to use the next line.
+  # By default the credential location is under $GCLOUD_LOCAL_DIR/credentials.
+  # Unless it is customized, there is no need to use the next line.
   #- $GOOGLE_APPLICATION_CREDENTIALS_LOCAL_DIR:/root/.secret
   - $FILESTORE_LOCAL_DIR:/app/__filestore
+  # This is useful for local development if we want changes to take effect
+  # immediately without restarting the docker service. Note that this will
+  # restart all services. If you want to narrow down the scope of the change,
+  # specify a desired folder like $LOCAL_PROJECT_ROOT/db:/app/db
+  - $LOCAL_PROJECT_ROOT:/app
```
ShahNewazKhan, line 15: @jzhang-gp since `docker-compose.yml` is in the root of the project we can use `$PWD:/app` here and same for `$FILESTORE_LOCAL_DIR` 

Note: If we use the `-f` flag then `PWD` will be where you issues the command from.

ShahNewazKhan (approved):

docker-compose.yml
jzhang-gp: @ShahNewazKhan Good point! I'll change to that instead :)

PR #145, opened 2020-10-07
Rename folders in the repo

jzhang-gp:
The Alchemy application has two customer facing components: the annotation server and the admin server. However, in the source code, they are named as `frontend` and `backend`, which can be confusing. 

This PR addresses the issue by explicitly naming the `frontend` to `annotation_server` and the `backend` to `admin_server` plus all related variable names and comments. **All changes here are just folder name and variable name changes except for the Dockerfile. No logic changes are included.**

The change in the Dockerfile is necessary since we need those dependencies to run integration tests. 

All existing unit and integration tests still pass.


Dockerfile
```diff
@@ -1,9 +1,21 @@
 FROM python:3.7-slim
 
+RUN apt-get update && apt-get install -y --no-install-recommends \
+	 build-essential \
+         wget \
+         curl \
+         git && \
```
ShahNewazKhan, line 7: I would move towards a multi stage build where only the compiled bin form the pip are carried to the final image similar to what's detailed here

https://www.rockyourcode.com/create-a-multi-stage-docker-build-for-python-flask-and-postgres/


Also the google recommended way of adding the sdk is 

```
# Install gcloud sdk & add to $PATH
RUN curl https://sdk.cloud.google.com | bash 
ENV PATH $PATH:/root/google-cloud-sdk/bin
```

ShahNewazKhan (approved):

Dockerfile
jzhang-gp: hmm changing the installation method somehow breaks the service. I'll keep it as is and add a task to investigate.

PR #147, opened 2020-10-14
Changes to resolve a bug on formatting the metrics

jzhang-gp:
This is a temporary fix for the bug in production on model metrics. Since the metrics can be `None`, rounding the number could cause error. 

The change here is just a quick patch on the code. In the long run, I want to move the computation of the metrics from the frontend to the backend as well as refactoring the metrics formatting in a better way plus good tests...

### DIFF
--- a/db/model.py
+++ b/db/model.py
@@ -3,12 +3,14 @@ import logging
 import copy
 import os
 import urllib.parse
+
+import numpy
 import pandas as pd
 import pickle
 from werkzeug.utils import secure_filename
 from flask_sqlalchemy import SQLAlchemy
 from sqlalchemy import create_engine, inspect, UniqueConstraint, MetaData, \
-    Boolean, distinct, desc
+    Boolean, desc, orm
 from sqlalchemy.schema import ForeignKey, Column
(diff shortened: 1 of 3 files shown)


db/model.py
```diff
@@ -949,3 +958,35 @@ def load_inference(version_dir: str, dataset_filename: str,
...
+
+    for metric in test_metrics:
+        test_metrics[metric] = _format_float_numbers(test_metrics[metric])
+
+    logging.info(metrics)
+
+    return metrics
+
+
+def _format_float_numbers(nums):
```
ShahNewazKhan, line 985: Given this is a utility function not tied to _reformat_metrics() should we also parameterize the precision and set default to `2f` ?

jzhang-gp: > Given this is a utility function not tied to `_reformat_metrics()` should we also parameterize the precision and set default to `2f` ?

Updated the precision to a parameter.

ShahNewazKhan: :+1:   :shipit:

ShahNewazKhan (approved):
:shipit:

PR #148, opened 2020-10-14
Weights are missing when the old votes are created.

jzhang-gp:
Hot patch code to fix a production bug where old votes do not have weights. They cause error with the new workflow that calculate priority based on weights. Unit tests are a bit tricky to add as it is part of a big function that needs refactoring. I'll add it in a separate PR.


ShahNewazKhan (approved):
LGTM!

PR #151, opened 2020-10-14
fix USE_CLOUD_LOGGING bug

### LINKED ISSUE #149: BUG: USE_CLOUD_LOGGING environment variable is ignored if a falsy value is provided
- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the latest release.
- [X] I have confirmed this bug exists on the main (master) branch.
-----
**Codesample**
```bash
USE_CLOUD_LOGGING=0 flask run
```

**Potential Source of the Problem**
Which is because the strings `'0'` or `'no'`, etc have truthy value, in here:
https://github.com/georgianpartners/annotation_tool/blob/5ad72faa5dc7f271b6e3d2d04bee003e1a6a6332/annotation_server/__init__.py#L17

**Potential Solution**
What `django-environ` does is to have a list of true values:

https://github.com/joke2k/django-environ/blob/027b2ef9d7000ce70aafd38b578ad1adb688d78b/environ/environ.py#L62

and checks the lower-cased value of boolean env variables against this list. A better solution would be to use [envparse](https://github.com/rconradharris/envparse), a framework agnostic port of `django-environ`.

MJafarMashhadi:
Closes #149

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ Descriptions of the env vars:
 - `ENV_STAGE`: Environment stage name.
 - `API_TOKEN_NAME`: The secret name for the inference API token.
 - `INFERENCE_OUTPUT_DATA_SOURCE_NAME_FOR_PUBSUB`: The inference output dataset source name for pubsub.
+- `USE_CLOUD_LOGGING`: Use GCP logging instead of the default logging. If set to true, `ADMIN_SERVER_LOGGGER` and `ANNOTATION_SERVER_LOGGER` must be set properly.
 
 New env vars:
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -92,3 +92,6 @@ vine==1.3.0
 wasabi==0.8.0
 Werkzeug==1.0.1
 zipp==3.1.0
+
+# Better environment variables 
+envparse==0.2.0
--- a/admin_server/__init__.py
+++ b/admin_server/__init__.py
@@ -1,6 +1,7 @@
 import os
 import logging
 
+from envparse import env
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -39,6 +39,7 @@ Descriptions of the env vars:
 - `ENV_STAGE`: Environment stage name.
 - `API_TOKEN_NAME`: The secret name for the inference API token.
 - `INFERENCE_OUTPUT_DATA_SOURCE_NAME_FOR_PUBSUB`: The inference output dataset source name for pubsub.
+- `USE_CLOUD_LOGGING`: Use GCP logging instead of the default logging. If set to true, `ADMIN_SERVER_LOGGGER` and `ANNOTATION_SERVER_LOGGER` must be set properly.
```
jzhang-gp, line 42: What would evaluate to True for this environment variable?

requirements.txt
```diff
@@ -92,3 +92,6 @@ vine==1.3.0
 wasabi==0.8.0
 Werkzeug==1.0.1
 zipp==3.1.0
+
+# Better environment variables 
+envparse==0.2.0
```
jzhang-gp, line 97: Have you rebuilt the images (except for the model training image) and checked the website is still working?

MJafarMashhadi: Just finished testing, it works. I also added the dependency explicitly in the Dockerfile as we discussed.

jzhang-gp:
LGTM!

jzhang-gp (approved):
LGTM

PR #153, opened 2020-10-15
Reorganizing folders under a root folder alchmey

jzhang-gp:
The current code structure is not ideal as they are scattered in multiple folders mixed with other non-source-code folders. It's very hard to find and track things that way. It also made it hard to setup code coverage reporting without a root source folder. 

**All changes here are folder name and hierarchy changes. There is no logic change.** 

All existing unit tests are passed.


ShahNewazKhan (approved):
LGTM!

PR #154, opened 2020-10-16
Using precommit-hook to reformat the code

jzhang-gp:
I'm trying to reformat the code with precommit-hook. These changes are sorted import statement and reformatted code. No logic changes.

I have to disable the `isort` module since it has conflicts with the `black` module and I like the style from the latter.


ShahNewazKhan (approved):
LGTM!

PR #155, opened 2020-10-22
Added google cloud build 

MJafarMashhadi:
branch name is misleading

### DIFF
--- a/alchemy/admin_server/external_services.py
+++ b/alchemy/admin_server/external_services.py
@@ -1,5 +1,8 @@
 import logging
 
+from google.cloud import pubsub
+from google.cloud import secretmanager
+
 
 class GCPPubSubService:
     _client = None
@@ -7,7 +10,6 @@ class GCPPubSubService:
     @classmethod
     def get_client(cls):
         if cls._client is None:
-            from google.cloud import pubsub
             cls._client = pubsub.PublisherClient()
         return cls._client
 
@@ -33,7 +35,6 @@ class SecretManagerService:
     @classmethod
     def get_client(cls):
         if cls._client is None:
-            from google.cloud import secretmanager
             cls._client = secretmanager.SecretManagerServiceClient()
         return cls._client
 
--- a/alchemy/admin_server/external_services.py
+++ b/alchemy/admin_server/external_services.py
@@ -1,16 +1,21 @@
 import logging
 
-from google.cloud import pubsub, secretmanager
-
 
 class GCPPubSubService:
-    publish_client = pubsub.PublisherClient()
+    _client = None
+
+    @classmethod
+    def get_client(cls):
+        if cls._client is None:
+            from google.cloud import pubsub
+            cls._client = pubsub.PublisherClient()
+        return cls._client
 
     @classmethod
     def publish_message(cls, project_id, topic_name, message_constructor, **kwargs):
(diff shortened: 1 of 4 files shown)


cloudbuild.yaml
```diff
@@ -0,0 +1,36 @@
...
+         '--file', 'Dockerfile',
+         '.']
+  id: 'step-build-base'
+
+# Step2: Build the production image by copying
+# the source code into the image
+#
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build',
+         '--tag', 'us.gcr.io/$PROJECT_ID/alchemy:latest',
```
MJafarMashhadi, line 18: Need to change it to `gcr.io`

alchemy/admin_server/external_services.py
```diff
@@ -1,16 +1,21 @@
...
-
 
 class GCPPubSubService:
-    publish_client = pubsub.PublisherClient()
+    _client = None
+
+    @classmethod
+    def get_client(cls):
+        if cls._client is None:
+            from google.cloud import pubsub
```
ShahNewazKhan, line 10: Moving the import to outside and at the top of the file will prevent from it being loaded each time the method is called.

cloudbuild.yaml
```diff
@@ -0,0 +1,36 @@
+steps:
+# Step1: Build the alchemy development image
+# that does not include the source codes as
+# the base
+#
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build',
+         '--tag', 'alchemy-base:latest',
```
ShahNewazKhan, line 8: We should use the git commit hash to tag the image as it makes it easier to debug later without having to generate the image digest and doing a reverse look up in the gcr. 

Cloud build provides both `$COMMIT_SHA` and `$SHORT_SHA` as detailed here https://cloud.google.com/cloud-build/docs/configuring-builds/substitute-variable-values

alchemy/admin_server/external_services.py
MJafarMashhadi: This is a lazy loading function, the import will be called only once in the if statement. Moved it to inside the method to have a cleaner namespace, however it's not a crucial part of my PR, I can move them back on top of the file if it's better.

cloudbuild.yaml
MJafarMashhadi: Thanks. The image label was a kind of grey area for me and I was looking forward to a feedback. So would a label like `dev-$SHORT_SHA` or `test-$SHORT_SHA` be good?

alchemy/admin_server/external_services.py
jzhang-gp: Generally I would prefer all the imports at the top so it's consistent and not scattered around from a readability perspective ( Java developer bias here ;) )

cloudbuild.yaml
ShahNewazKhan: No worries! Generally the idea is to build one image to be used through out all environments so just `$SHORT_SHA` is good.

MJafarMashhadi: Makes sense, Thanks!

ShahNewazKhan (approved):
LGTM!

jzhang-gp (approved):
LGTM

PR #157, opened 2020-10-22
Separated local, production, and testing requirements files.

MJafarMashhadi:


Dockerfile
```diff
@@ -10,6 +10,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 WORKDIR /base
 
 COPY requirements.txt .
+COPY requirements ./requirements
 RUN pip install -r requirements.txt
```
jzhang-gp, line 14: Hmm I'm a bit confused if we are using the `requirements` folder, do we still need to copy the requirement.txt? Should we also install one of the dependency file under the `requirements` folder?

MJafarMashhadi: That is just a reference to production requirements. I put it there just in case any external tools, now or in future, rely on the requirements file to be in the root directory. (Heroku is an example)

jzhang-gp: > That is just a reference to production requirements. I put it there just in case any external tools, now or in future, rely on the requirements file to be in the root directory. (Heroku is an example)

Can you elaborate on that a bit? Why do we need the file to be in the root directory for external tools?

MJafarMashhadi: Is the question on why I copied it in the Docker file or why it exists?

MJafarMashhadi: So for example if a portfolio company wants to deploy a copy of Alchemy on Heroku, it expects a requirements.txt in the root of the repository. That's a reason why I kept it there.

jzhang-gp: Hmm I'm not sure I follow...can we do this instead if the `requirements.txt` is only referencing the `production.txt`? 
```
COPY requirements ./requirements
RUN pip install -r ./requirements/production.txt
```

MJafarMashhadi: Yup, will change it

jzhang-gp (approved):
LGTM

PR #158, opened 2020-10-23
Cloud build training image

MJafarMashhadi:
Created a trigger on GCP that only builds the training image if relevant files are changed in a commit to main branch.

### DIFF
--- a/alchemy/train/no_deps/cloudbuild.yaml
+++ b/alchemy/train/no_deps/cloudbuild.yaml
@@ -1,3 +1,9 @@
+# This build file is triggered on each commit to master
+# in which some files in alchemy/train/nodeps/** are
+# touched.
+# This is configured in the GCP console > cloud build >
+# triggers. See:
+# https://cloud.google.com/cloud-build/docs/automating-builds/create-github-app-triggers#creating_github_app_triggers_2
 steps:
 - name: 'gcr.io/cloud-builders/docker'
   args: ['build',
--- a/alchemy/train/no_deps/cloudbuild.yaml
+++ b/alchemy/train/no_deps/cloudbuild.yaml
@@ -1,0 +1,9 @@
+steps:
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build',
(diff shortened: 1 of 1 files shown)


jzhang-gp:
How does this cloudbuild file alone detects that the changes are for the training image?

MJafarMashhadi:
Set it to trigger if both conditions are true:
- push is on a branch matching `/^master$/`
- a change in `alchemy/train/no_deps/**` is detected

jzhang-gp:
> Set it to trigger if both conditions are true:
> 
>     * push is on a branch matching `/^master$/`
> 
>     * a change in `alchemy/train/no_deps/**` is detected

So the trigger is on Cloud build console, correct? There is no code changes involving that, right?

MJafarMashhadi:
> > Set it to trigger if both conditions are true:
> > ```
> > * push is on a branch matching `/^master$/`
> > 
> > * a change in `alchemy/train/no_deps/**` is detected
> > ```
> 
> So the trigger is on Cloud build console, correct? There is no code changes involving that, right?

Correct. Referring to the comment on #159 , it would be better to have that config it in the codebase itself

jzhang-gp:
> > > Set it to trigger if both conditions are true:
> > > ```
> > > * push is on a branch matching `/^master$/`
> > > 
> > > * a change in `alchemy/train/no_deps/**` is detected
> > > ```
> > 
> > 
> > So the trigger is on Cloud build console, correct? There is no code changes involving that, right?
> 
> Correct. Referring to the comment on #159 , it would be better to have that config it in the codebase itself

Yeah that make sense. Could you add a comment in the cloud build file to explain the conditions and they are currently specified in the cloud build console?

jzhang-gp (approved):
As mentioned in the discussion, let's temporarily add a comment to explain the conditions and where they are specified.

MJafarMashhadi:
> As mentioned in the discussion, let's temporarily add a comment to explain the conditions and where they are specified.

Added, also added some details in confluence page "CI/CD Pipeline Roadmap"

PR #159, opened 2020-10-23
Updated build file to use TARGET for being able to build multiple images 

MJafarMashhadi:
Fixes: CU-epv2kw 

Updates:
1. It now builds a testing image and run tests on that
2. After tests pass, it will build the target image
3. Image tags now include the target as well. e.g. `test-62ba7cc`, `production-a335237`, `staging-172b61c`, etc.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,36 +1,55 @@
 steps:
-# Step1: Build the alchemy development image
-# that does not include the source codes as
-# the base
+# Build the alchemy test image
 #
 - name: 'gcr.io/cloud-builders/docker'
   args: ['build',
-         '--tag', 'alchemy-base:$SHORT_SHA',
+         '--build-arg', 'TARGET=test',
+         '--tag', 'alchemy-base',
          '--file', 'Dockerfile',
          '.']
-  id: 'step-build-base'
-
-# Step2: Build the production image by copying
-# the source code into the image
-#
+  id: 'build-test-image-base'
 - name: 'gcr.io/cloud-builders/docker'
   args: ['build',
-         '--tag', 'gcr.io/$PROJECT_ID/alchemy:$SHORT_SHA',
+         '--build-arg', 'TARGET=test',
+         '--tag', 'gcr.io/$PROJECT_ID/alchemy:test-${SHORT_SHA}',
          '--file', 'alchemy.Dockerfile',
          '.']
-  wait_for: ['step-build-base']
-  id: 'step-build-prod'
+  wait_for: ['build-test-image-base']
+  id: 'build-test-image-full'
 
-# Step3: Run the tests
+# Run the tests, to prevent having broken
+# images. May look unnecessary since there are
+# GitHub actions that run tests, however it
+# wouldn't hurt to double check.
 #
 - name: 'gcr.io/cloud-builders/docker'
   args: ['run',
          '--env', 'GOOGLE_AI_PLATFORM_ENABLED=0',
          '--env', 'USE_CLOUD_LOGGING=0',
-         'gcr.io/$PROJECT_ID/alchemy:$SHORT_SHA',
+         'gcr.io/$PROJECT_ID/alchemy:test-${SHORT_SHA}',
          'sh', 'ci/run_tests.sh']
-  wait_for: ['step-build-prod']
-  id: 'step-test'
-timeout: 600s
+  wait_for: ['build-test-image-full']
+  id: 'run-tests'
+
+# Build the target image
+#
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build',
+         '--build-arg', 'TARGET=${_TARGET}',
+         '--tag', 'alchemy-base',
+         '--file', 'Dockerfile',
+         '.']
+  wait_for: ['run-tests']
+  id: 'build-target-image-base'
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build',
+         '--build-arg', 'TARGET=${_TARGET}',
+         '--tag', 'gcr.io/$PROJECT_ID/alchemy:${_TARGET}-${SHORT_SHA}',
+         '--file', 'alchemy.Dockerfile',
+         '.']
+  wait_for: ['build-target-image-base']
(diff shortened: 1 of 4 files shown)


MJafarMashhadi:
I have set up the GCP build triggers on the GCP console. But I think it would have been nicer to have some files in the repository that configure the GCP. I know that there are some solutions; configuration as code or something? What are your thoughts?

jzhang-gp:
> I have set up the GCP build triggers on the GCP console. But I think it would have been nicer to have some files in the repository that configure the GCP. I know that there are some solutions; configuration as code or something? What are your thoughts?

That sounds like something Terraform can do?

MJafarMashhadi:
> > I have set up the GCP build triggers on the GCP console. But I think it would have been nicer to have some files in the repository that configure the GCP. I know that there are some solutions; configuration as code or something? What are your thoughts?
> 
> That sounds like something Terraform can do?

Probably, it makes sense for Terraform to do it. I have not read enough about it yet.

cloudbuild.yaml
```diff
@@ -1,36 +1,55 @@
...
-  id: 'step-test'
-timeout: 600s
+  wait_for: ['build-test-image-full']
+  id: 'run-tests'
+
+# Build the target image
+#
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build',
+         '--build-arg', 'TARGET=${_TARGET}',
```
jzhang-gp, line 38: How is the value of `_TARGET` supplied into cloud build? In the console?

cloudbuild.yaml
```diff
@@ -1,36 +1,55 @@
 steps:
-# Step1: Build the alchemy development image
-# that does not include the source codes as
-# the base
+# Build the alchemy test image
 #
 - name: 'gcr.io/cloud-builders/docker'
   args: ['build',
-         '--tag', 'alchemy-base:$SHORT_SHA',
+         '--build-arg', 'TARGET=test',
```
jzhang-gp, line 6: So the `--build-arg` and `TARGET=test` here will override the `ARG TARGET=production` in the Dockerfile?

cloudbuild.yaml
MJafarMashhadi: Yes, it can be both provided in the console (for triggers) which I hard-coded to `production` since it gets triggered on `master` branch commits and in the command line. For testing it I submitted the job like this:

```bash
gcloud builds submit --substitutions=SHORT_SHA=$(git rev-parse --short $(git log -1 --pretty=format:%h)),_TARGET=production .
```

cloudbuild.yaml
MJafarMashhadi: It builds the image two times, once with test configuration, for running tests and once with `_TARGET` config after tests have passed. The only image that is pushed to the registry is the target image. The test image will be discarded. For now, I have hard-coded the target to be `production` in GCP console. Once the pipeline is working we can add more configurations such as nightly builds.

MJafarMashhadi: > So the `--build-arg` and `TARGET=test` here will override the `ARG TARGET=production` in the Dockerfile?

Oh, I misunderstood the question. Yes it overrides the one in Dockerfile.

jzhang-gp (approved):
LGTM. Can you document the instructions on how to use the cloud build commands? Thanks.

MJafarMashhadi:
> LGTM. Can you document the instructions on how to use the cloud build commands? Thanks.

Sure, I'll add them in Confluence.

cloudbuild.yaml
ShahNewazKhan: Does it make sense to build the image once and over-ride the entrypoint for the different workflows, it's best practice to only have one artifact used across all environments. 

Here is an example we use in the data-platfrom to build once and over-ride the entrypoint to run tests. 

https://github.com/georgianpartners/data_platform/blob/beta/cloudbuild-dags.yaml#L27-L49

MJafarMashhadi: I don't have permissions to see that file, can you send it to me on Slack or email?
Also, the differences can be in the dependencies as well, not only the entry point. For example the testing image has pytest, mockito, etc while tge production image doesn't have them. Or local development image might use a debugging webserver while tge prod image uses gunicorn for example. How can we handle that?

ShahNewazKhan: Currently we are doing a multi-stage `Dockerfile` with each stage holding only the build dependencies for that environment. This way only the stage based binaries are present and can be used with the `--target` docker param as describe [here](https://www.docker.com/blog/speed-up-your-development-flow-with-these-dockerfile-best-practices/)  

Here is the `cloud-build` code snippet I was referring to:

```
# build & push core image
- name: 'gcr.io/cloud-builders/docker'
  entrypoint: 'bash'
  args: ['-c', 'docker pull gcr.io/georgian-data-platform/core:latest || exit 0']

- name: 'gcr.io/cloud-builders/docker'
  args: [ 
    'build', 
    '-t', 'gcr.io/georgian-data-platform/core:$SHORT_SHA', 
    '-t', 'gcr.io/georgian-data-platform/core:latest', 
    '--cache-from', 'gcr.io/georgian-data-platform/core:latest',
    '.' ]
  dir: 'wranglers/core'
- name: 'gcr.io/cloud-builders/docker'
  args: ['push', 'gcr.io/georgian-data-platform/core:$SHORT_SHA']
- name: 'gcr.io/cloud-builders/docker'
  args: ['push', 'gcr.io/georgian-data-platform/core:latest']

# run wrangler unit tests 
- name: 'gcr.io/georgian-data-platform/core:$SHORT_SHA'
  args: [ "sh","-c","/opt/venv_core/bin/pytest /app/core_utils/tests --disable-pytest-warnings"]
  env:
    - 'PYTHONPATH=/app/core_utils'
```

MJafarMashhadi: I was just reading about multi-stage builds at the same time that you posted this comment, I needed it for another task. With the link you shared it's more clear now. It's brilliant! I've updated the file according to this review; could you please review the new changes?

ShahNewazKhan (approved):
LGTM!

PR #162, opened 2020-10-28
Fixing the python import problem without using the sys path hack

jzhang-gp:
One of the precommit-hook changed the order of some import statement in the `alembic` tool. This messed up the code since the code was using the `sys.path = ["", ".."] + sys.path[1:]` hack to find a module in a different directory. This is not  ideal. 

According to [this stackoverflow post](https://stackoverflow.com/questions/49039436/how-to-import-a-module-from-a-different-folder), we can solve the import path issue by setting the PYTHONPATH to the root folder and hence the change here. 

All tests pass and the migration scripts work too.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,17 +17,17 @@ RUN sh ci/install_deps.sh
 
 # Path configuration
 ENV PATH $PATH:/root/tools/google-cloud-sdk/bin
+# This is necessary for the alembic migration to work
+ENV PYTHONPATH /app
 
 FROM alchemy-base as local
 RUN pip install -r requirements/local.txt
 WORKDIR /app
-ENV PYTHONPATH /app
 
 
 FROM alchemy-base as alchemy-with-code
 WORKDIR /app
 COPY . .
-ENV PYTHONPATH /app
 
 FROM alchemy-with-code as production
 RUN pip install -r requirements/production.txt
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,11 +21,13 @@ ENV PATH $PATH:/root/tools/google-cloud-sdk/bin
 FROM alchemy-base as local
 RUN pip install -r requirements/local.txt
 WORKDIR /app
+ENV PYTHONPATH /app
 
 
 FROM alchemy-base as alchemy-with-code
 WORKDIR /app
 COPY . .
+ENV PYTHONPATH /app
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -21,11 +21,13 @@ ENV PATH $PATH:/root/tools/google-cloud-sdk/bin
 FROM alchemy-base as local
 RUN pip install -r requirements/local.txt
 WORKDIR /app
+ENV PYTHONPATH /app
 
 
 FROM alchemy-base as alchemy-with-code
 WORKDIR /app
 COPY . .
+ENV PYTHONPATH /app
```
MJafarMashhadi, line 30: Could we define it only once in line 20 in `alchemy-base`?

jzhang-gp: Done.

MJafarMashhadi:
LGTM

ShahNewazKhan (approved):
It will be good to note this update in a `README.md` type section as setting the `PYTHONPATH` changes the order in which modules are resolved during `imports` and if you have a name collision with another depencencies it makes debugging a nightmare. 

Otherwise LGTM!

jzhang-gp:
> It will be good to note this update in a `README.md` type section as setting the `PYTHONPATH` changes the order in which modules are resolved during `imports` and if you have a name collision with another depencencies it makes debugging a nightmare.
> 
> Otherwise LGTM!

Hey Shah, is there an alternative to fix this problem? I could make a change to the pre-commit hook to exclude the `env.py` file but I'm not sure if that's a clean solution.

ShahNewazKhan:
> > It will be good to note this update in a `README.md` type section as setting the `PYTHONPATH` changes the order in which modules are resolved during `imports` and if you have a name collision with another depencencies it makes debugging a nightmare.
> > Otherwise LGTM!
> 
> Hey Shah, is there an alternative to fix this problem? I could make a change to the pre-commit hook to exclude the `env.py` file but I'm not sure if that's a clean solution.

@jzhang-gp I think the cleanest way to implement this is by using `PYTHONPATH` as that is what that envar is intended to do, however in this case we are setting this envar for the all the workflows in the `docker image`. 

I think we can reduce the scope of the `PYTHONPATH` by prepending it where where we run `alembic` cmds as such 

```
PYTHONPATH=/app python alembic upgrade 
```

PR #165, opened 2020-10-28
Restructured the configurations to be read from the `app.config` instead of being scattered around

MJafarMashhadi:


MJafarMashhadi:
No longer needed / moving to Django

PR #152, opened 2020-10-15
Refactored environment variables using envparse

MJafarMashhadi:
closes #150 
xref #151 , #150

PR #166, opened 2020-10-29
Added envparse to the requirements.txt of the cloud funtion

MJafarMashhadi:
xref #152


jzhang-gp (approved):
LGTM. Could you add a clickup task of deploying this function in production so we don't forget to update it? Hopefully in the future we can also deploy it through the cloud build pipeline.

MJafarMashhadi:
> LGTM. Could you add a clickup task of deploying this function in production so we don't forget to update it? Hopefully in the future we can also deploy it through the cloud build pipeline.

Thanks, just adding it.

PR #167, opened 2020-10-29
Annotation DAO Initial Version

jzhang-gp:
This is the first attempt to extract some of the Upsert logic away from the controller into a DAO layer. Basically people can bulk upload annotations for data labeling task. The uploaded data may involve both new and existing entities. 

I have some thoughts on making the application more service-oriented given that we may have other clients creating annotations through API calls. The `AnnotationUpsertRequest` is created for that purpose. There are more work to be done for the service itself and will be added in future PRs.

### DIFF
--- a/alchemy/dao/annotation_dao.py
+++ b/alchemy/dao/annotation_dao.py
@@ -1,24 +1,7 @@
 from typing import List
 
 from alchemy.data.request.annotation_request import AnnotationUpsertRequest
-from alchemy.db.model import ClassificationAnnotation, EntityTypeEnum
-
-
-# TODO we should move this to a utility class in case we have more than one
-#  entity_type in the future
-def _construct_context(entity_type, entity):
-    if entity_type == EntityTypeEnum.COMPANY:
-        context = {"text": "N/A", "meta": {"name": entity, "domain": entity}}
-    else:
-        context = {
-            "text": "N/A",
-            "meta": {
-                "name": entity
-                # TODO we probably should name `domain` to
-                #  something else according to the entity type
-            },
-        }
-    return context
+from alchemy.db.model import ClassificationAnnotation
 
 
 class AnnotationDao:
@@ -57,24 +40,20 @@ class AnnotationDao:
             upsert_request.user_id,
         )
         if annotation is None:
-            context = _construct_context(
-                upsert_request.entity_type, upsert_request.entity
-            )
             annotation = ClassificationAnnotation(
                 entity_type=upsert_request.entity_type,
                 entity=upsert_request.entity,
                 user_id=upsert_request.user_id,
                 label=upsert_request.label,
                 value=upsert_request.value,
-                context=context,
+                context=upsert_request.context,
             )
         else:
-            # Normally only the annotation value should be updated.
-            # annotation.entity_type = upsert_request.entity_type
-            # annotation.entity = upsert_request.entity
-            # annotation.user_id = upsert_request.user_id
-            # annotation.label = upsert_request.label
-            # annotation.context = upsert_request.context
+            annotation.entity_type = upsert_request.entity_type
+            annotation.entity = upsert_request.entity
+            annotation.user_id = upsert_request.user_id
+            annotation.label = upsert_request.label
+            annotation.context = upsert_request.context
             annotation.value = upsert_request.value
         return annotation
 
--- a/alchemy/dao/annotation_dao.py
+++ b/alchemy/dao/annotation_dao.py
@@ -1,19 +1,40 @@
 from typing import List
 
+from sqlalchemy.exc import DatabaseError
+
 from alchemy.data.request.annotation_request import AnnotationUpsertRequest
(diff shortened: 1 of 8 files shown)


alchemy/dao/annotation_dao.py
```diff
@@ -0,0 +1,65 @@
...
+        annotation = (
+            self.dbsession.query(ClassificationAnnotation)
+            .filter_by(
+                entity_type=entity_type, entity=entity, user_id=user_id, label=label
+            )
+            .one_or_none()
+        )
+        return annotation
+
+    def _upsert_annotation_helper(self, upsert_request: AnnotationUpsertRequest):
```
ShahNewazKhan, line 35: @jzhang-gp if we have concurrent runs of this upsert check, would the 2 commit be rejected for the case where the document didn't exist?

jzhang-gp: In theory, it's possible. If there is high traffic on the API, this could become a problem. Our use case has a very low call rate. This API is only used by the website to upload a batch of documents by the annotator manually. So there is less risk. We could also add some retry logic. Do you think that would help with the issue? Any other suggestions?

alchemy/data/request/base_request.py
```diff
@@ -0,0 +1,21 @@
+class InvalidRequest:
```
MJafarMashhadi, line 1: Suggestion: either
1. Adding a request base class that both of these inherit from
2. Combining the two, adding an `is_valid` method/property
```
@property
def is_valid():
   if self.errors is None:
       self._validate()
   return not self.has_errors()

def _validate(self):
    # ... do some generic validations...
    # then:
    child_validators = [getattr(self, validator) for validator in dir(self) if validator.startswith('validate_') or validator == 'validate']
    for c in child_validators:
        if callable(c):
            c() # or maybe use the return value
```
Then in a child class, such as `AnnotationUpsertRequest` we can have:

```
class AnnotationUpsertRequest(Request):
  def validate(self):
    # custom validations
    pass

  def validate_user_id(self):
    # moar cusom validation
    pass
```

alchemy/data/request/annotation_request.py
```diff
@@ -0,0 +1,45 @@
...
+class AnnotationUpsertRequest(ValidRequest):
+    entity_type: str
+    entity: str
+    label: str
+    user_id: int
+    value: int
+    context: dict
+
+    @classmethod
+    def from_dict(cls, dict_data):
```
MJafarMashhadi, line 16: What about removing this function and using `__post_init__` to perform the validations?

jzhang-gp: I initially had code like the following but decided to remove it. 
```
    def __post_init__(self):
        dict_data = as_dict(self)
        self._validate_request_data(dict_data=dict_data,
                                    invalid_req=InvalidRequest())
```
The current way to create the UpsertRequests are rather temporary. In the deal case, we are going to be handle REST API calls with JSON payload, which we can process with `AnnotationUpsertRequest.from_dict`. 

But I can add it until the service is up and running.

alchemy/data/request/base_request.py
jzhang-gp: Added a base class `Request` which is empty:
```
class Request:
    pass
```

alchemy/dao/annotation_dao.py
ShahNewazKhan: @jzhang-gp that makes sense, in case of low traffic, there is probably even lesser chance of concurrent transactions updating the same row. 

In the case of high traffic we can use [sql alchemy's row versioning](https://docs.sqlalchemy.org/en/13/orm/versioning.html), however this requires a data migration.

MJafarMashhadi (approved):

alchemy/dao/annotation_dao.py
jzhang-gp: Added some retry logic. I'll also look into the row versioning. Thanks @ShahNewazKhan

alchemy/dao/annotation_dao.py
```diff
@@ -1,19 +1,40 @@
...
+
 
 class AnnotationDao:
     def __init__(self, dbsession):
         self.dbsession = dbsession
 
+    @retry(
+        stop=stop_after_attempt(3),
+        wait=wait_exponential(multiplier=1, min=2, max=6),
+        retry=retry_if_exception_type(DatabaseError),
```
ShahNewazKhan, line 23: Should this also cover [DBAPI](https://docs.sqlalchemy.org/en/13/core/exceptions.html#sqlalchemy.exc.DBAPIError) error?

jzhang-gp: > Should this also cover [DBAPI](https://docs.sqlalchemy.org/en/13/core/exceptions.html#sqlalchemy.exc.DBAPIError) error?

Good point. I'll update it.

ShahNewazKhan (approved):
LGTM!

PR #168, opened 2020-10-30
Cloud build improvements

MJafarMashhadi:
* Disabling pip caches
* Smarter build pipeline
* Tagging 'production' images as latest

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 WORKDIR /base
 
 COPY requirements ./requirements
-RUN pip install -r requirements/base.txt
+RUN pip install --no-cache -r requirements/base.txt
 
 COPY ci ./ci
 RUN sh ci/install_deps.sh
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -22,14 +22,22 @@ steps:
   wait_for: ['build-test-image']
   id: 'run-tests'
 
-# Build the target image
+# Build the target image (if target != test)
 #
 - name: 'gcr.io/cloud-builders/docker'
-  args: ['build',
-         '--target', '${_TARGET}',
-         '--cache-from', 'gcr.io/$PROJECT_ID/alchemy:test-${SHORT_SHA}',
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -10,7 +10,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 WORKDIR /base
 
 COPY requirements ./requirements
-RUN pip install -r requirements/base.txt
+RUN pip install --no-cache -r requirements/base.txt
```
jzhang-gp, line 13: What effect would `--no-cache` have on the build?

cloudbuild.yaml
```diff
@@ -22,17 +22,43 @@ steps:
   wait_for: ['build-test-image']
   id: 'run-tests'
 
-# Build the target image
+# Build the target image (if target != test)
```
jzhang-gp, line 25: Why do we skip the build for test?

Dockerfile
MJafarMashhadi: The build itself will be the same, but the size of the image will be smaller because pip won't cache the whl and other intermediate files. The cache is useful on a personal computer but not in a docker image, It unnecessarily increases the size.

cloudbuild.yaml
MJafarMashhadi: Since we build the test image every time (no matter what the target is) if the final target is "test" too we don't need to build the exact same image twice, so I just skip it.

jzhang-gp (approved):
LGTM

MJafarMashhadi:
Thanks!

MJafarMashhadi:
@ShahNewazKhan Hey could you please have a quick review of this PR? I ask for your review you on PRs that are related to GCP

Dockerfile
ShahNewazKhan: This makes sense, however since we are using multi-stage builds. the `--target` flag allows the image build process to stop at a specific stage. This allows us to build a `local` image vs a `prod` image based on the build stage the `--target` specifies. 

If the image size is a concern, we should copy over the binaries from the pip install to the next stage and prevent the build tools from cascading down the build stages. 

This way we can leverage the image caching feature in `cloud build` to speed up our builds as well as reduce the size of the image.

ShahNewazKhan (approved):
LGTM!

PR #169, opened 2020-11-02
Add a proper entry point for the docker image

MJafarMashhadi:


Dockerfile
```diff
@@ -30,6 +33,44 @@ WORKDIR /app
...
+    && apt-get purge -y --auto-remove $buildDeps \
+    && find /usr/local -depth \
+    \( \
+        \( -type d -a -name test -o -name tests \) \
+        -o \
+        \( -type f -a -name '*.pyc' -o -name '*.pyo' \) \
+    \) -exec rm -rf '{}' +
+
+FROM alchemy-with-code as staging
 RUN pip install -r requirements/production.txt
```
jzhang-gp, line 74: I'm a bit confused. Are we not installing the production requirements into the `production` stage?

Dockerfile
```diff
@@ -20,6 +20,9 @@ ENV PATH $PATH:/root/tools/google-cloud-sdk/bin
 # This is necessary for the alembic migration to work
 ENV PYTHONPATH /app
 
+ENTRYPOINT ["ci/run_server.sh"]
+CMD ["admin_server"]
```
jzhang-gp, line 24: Is the default parameter necessary?

jzhang-gp (commented):
Have you tested the server is up and running (create a task, do a few annotations and train a model) since this involves changes on how we run production server?

Dockerfile
MJafarMashhadi: The CMD line is not very necessary; the ENTRYPOINT command should be there though

Dockerfile
MJafarMashhadi: Yes, we are installing production requirements in production stage.
uwsgi needs a C compiler to install, these commands install `gcc`, `make`, etc before running `pip install -r requirements/production.txt` and after that, it removes them all.

MJafarMashhadi:
> Have you tested the server is up and running (create a task, do a few annotations and train a model) since this involves changes on how we run production server?

Yes, it works :)

MJafarMashhadi:
> Have you tested the server is up and running (create a task, do a few annotations and train a model) since this involves changes on how we run production server?

It works, yes

Dockerfile
jzhang-gp: Ah I see. It's mixed with the rest of the command.

Dockerfile
jzhang-gp: Can we remove it? I feel this is a little bit confusing.

jzhang-gp (approved):
Just a minor comment on the `CMD` in the dockerfile. The rest looks fine.

Dockerfile
MJafarMashhadi: Done :)

PR #170, opened 2020-11-05
Guard against duplicate labels when a Task object is created. 

jzhang-gp:
In Alchemy, an annotation task can include multiple labels for annotators to work on. For example, in one task, we ask annotators to annotate companies into either "B2C" or "Machine Learning". 

Labels should not be reused in different tasks. Unfortunately given the current design, label is neither treated as its own table nor a part of a primary key. They are treated as strings in a JSON column. 

As a temporary solution, we are checking the duplicates when creating a task. There will be another PR for this check in task update as well.

### DIFF
--- a/alchemy/admin_server/tasks.py
+++ b/alchemy/admin_server/tasks.py
@@ -6,37 +6,64 @@ import pandas as pd
 
 from envparse import env
 from flask import (
-    Blueprint, flash, redirect, render_template, request, url_for, send_file
+    Blueprint,
+    flash,
+    redirect,
+    render_template,
+    request,
+    url_for,
+    send_file,
 )
 from werkzeug.utils import secure_filename
 
+from alchemy.data.request.task_request import TaskCreateRequest
 from alchemy.db.model import (
-    db, ClassificationAnnotation, User, Task, Model, AnnotationGuide,
-    LabelPatterns, ModelDeploymentConfig, EntityTypeEnum,
-    delete_requests_for_user_under_task, delete_requests_for_label_under_task,
-    delete_requests_under_task, delete_requests_for_entity_type_under_task
+    db,
+    ClassificationAnnotation,
+    User,
+    Task,
+    Model,
+    AnnotationGuide,
+    LabelPatterns,
+    ModelDeploymentConfig,
+    EntityTypeEnum,
+    delete_requests_for_user_under_task,
+    delete_requests_for_label_under_task,
+    delete_requests_under_task,
+    delete_requests_for_entity_type_under_task,
 )
 from alchemy.db.utils import get_all_data_files
 
 from alchemy.ar.data import (
-    compute_annotation_statistics_db, compute_annotation_request_statistics
+    compute_annotation_statistics_db,
+    compute_annotation_request_statistics,
 )
 from alchemy.ar.ar_celery import generate_annotation_requests
 
 from alchemy.train.train_celery import submit_gcp_training
 
 from alchemy.shared.celery_job_status import (
-    CeleryJobStatus, create_status, delete_status
+    CeleryJobStatus,
+    create_status,
+    delete_status,
 )
 from alchemy.shared.annotation_server_path_finder import (
-    generate_annotation_server_user_login_link, generate_annotation_server_admin_examine_link,
-    generate_annotation_server_compare_link
+    generate_annotation_server_user_login_link,
+    generate_annotation_server_admin_examine_link,
+    generate_annotation_server_compare_link,
 )
 from alchemy.shared.utils import (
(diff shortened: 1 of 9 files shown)


alchemy/dao/task_dao.py
```diff
@@ -0,0 +1,75 @@
...
+        """This is a temporary hacky solution to prevent users from creating
+        duplicate labels. The proper solution involves a much bigger
+        refactoring.
+
+        Given the amount of tasks and labels we have, the performance is still
+        acceptable.
+
+        TODO replace this once we have the Market Management System in place.
+        """
+        tasks = self.dbsession.query(Task).all()
```
ShahNewazKhan, line 53: Does it make sense to leverage the db to do the filtering for example using the `in_` https://docs.sqlalchemy.org/en/14/core/sqlelement.html#sqlalchemy.sql.expression.ColumnOperators.in_ filter to see if the label exists?

alchemy/dao/task_dao.py
```diff
@@ -0,0 +1,75 @@
...
+        stop=stop_after_attempt(3),
+        wait=wait_exponential(multiplier=1, min=2, max=6),
+        retry=retry_if_exception_type(DBAPIError),
+        reraise=True,
+    )
+    def create_task(self, create_request):
+        if not create_request:
+            # TODO We should consider adding Response Object in pair with the
+            #  Request Object.
+            raise Exception(f"Invalid request: {create_request.errors}")
```
MJafarMashhadi, line 26: Maybe `ValueError` would be a better exception type to raise

alchemy/dao/task_dao.py
```diff
@@ -0,0 +1,75 @@
...
+                existing_labels[label] = task.name
+        error_msgs = []
+        for new_label in new_labels:
+            if new_label in existing_labels:
+                error_msgs.append(
+                    f"Label {new_label} is already created "
+                    f"in task {existing_labels[new_label]}"
+                )
+        if len(error_msgs) > 0:
+            raise Exception(error_msgs)
```
MJafarMashhadi, line 68: Similarly, `ValueError` could be a more specific exception type to raise:
https://docs.python.org/3/library/exceptions.html#exception-hierarchy

MJafarMashhadi (approved):
Other than theLGTM

alchemy/admin_server/tasks.py
```diff
@@ -78,14 +105,18 @@ def create():
         annotators = parse_annotators(form)
         data_files = parse_data(form, data_fnames)
 
-        task = Task(name=name)
-        task.set_labels(labels)
-        task.set_annotators(annotators)
-        task.set_data_filenames(data_files)
-        task.set_entity_type(entity_type)
+        print("Before creating the request!")
```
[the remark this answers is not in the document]
jzhang-gp: Ah yes I'll remove them.

alchemy/dao/task_dao.py
jzhang-gp: It might be tricky because the labels are nested in a JSON column but I'll give it a short. Thanks for the suggestion!

alchemy/admin_server/tasks.py
```diff
@@ -260,9 +291,9 @@ def update(id):
 
 @bp.route("/<string:id>/assign", methods=["POST"])
 def assign(id):
-    max_per_annotator = env.int('ANNOTATION_TOOL_MAX_PER_ANNOTATOR', default=100)
-    max_per_dp = env.int('ANNOTATION_TOOL_MAX_PER_DP', default=3)
-    entity_type = request.form.get('entity_type')
+    max_per_annotator = env.int("ANNOTATION_TOOL_MAX_PER_ANNOTATOR", default=100)
```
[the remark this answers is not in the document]
jzhang-gp: I can add this to backlog :)

alchemy/dao/task_dao.py
jzhang-gp: I have something else in mind on this one. I'll send a separate PR for this :)

alchemy/dao/task_dao.py
jzhang-gp: Same as above.

alchemy/dao/task_dao.py
ShahNewazKhan: I agree, you can use the `astext` and `_in` filter similar to [this](https://stackoverflow.com/a/53266167/2990066)

jzhang-gp: Looked into this. The labels are saved as a list inside a JSON column. `IN` doesn't seem to support checking `a list in another list`. It might be possible to do this with some hacks but I feel the readability would suffer. I'm going to stick to this temporary solution for now.

PR #172, opened 2020-11-06
Automatic deployments to compute engine

MJafarMashhadi:
TOFF: A temporary way to have auto deploy on a managed VM, with ssh and docker-compose.
Next steps are deploying on cloud run and GKE.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -68,35 +68,17 @@ steps:
     - "-c"
     - |
       if [ "${_TARGET}" = "production" ] && [ -z ${_NO_DEPLOY} ]; then
-        echo Downloading the keys
-        gcloud secrets versions access latest --secret=alchemy-builder-ssh > id_rsa
-        gcloud secrets versions access latest --secret=alchemy-builder-ssh-pub > id_rsa.pub
-        echo "" >> id_rsa ; echo "" >> id_rsa.pub ;  # Add a new line to the end
-
         # INSTANCES=$(gcloud compute instances list --filter="${_DEPLOY_FILTER}")
-        echo Copying the keys
-        gcloud compute ssh \
-            --zone "us-central1-a" \
-            --project "nlp-flywheel" \
-            "${_DEPLOY_MACHINE}" -- "mkdir -p ~/.ssh/"
-        gcloud compute scp \
-            --zone "us-central1-a" \
-            --project "nlp-flywheel" \
-            id_rsa* ${_DEPLOY_MACHINE}:~/.ssh/
-
         echo "Updating the project"
         gcloud compute ssh \
             --zone "us-central1-a" \
             --project "nlp-flywheel" \
             "${_DEPLOY_MACHINE}" -- " \
-              chmod 600 ~/.ssh/id_rsa
-              chmod 644 ~/.ssh/id_rsa.pub
               cd /var/www
               git pull
               docker pull gcr.io/nlp-flywheel/alchemy:latest && \
               docker-compose -f docker-compose-staging.yml down && \
-              docker-compose -f docker-compose-staging.yml up -d
-              rm -rf ~/.ssh/"
+              docker-compose -f docker-compose-staging.yml up -d"
       else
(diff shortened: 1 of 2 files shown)


cloudbuild.yaml
```diff
@@ -59,6 +59,47 @@ steps:
...
+        gcloud compute ssh \
+            --zone "us-central1-a" \
+            --project "nlp-flywheel" \
+            "${_DEPLOY_MACHINE}" -- " \
+              chmod 600 ~/.ssh/id_rsa
+              chmod 644 ~/.ssh/id_rsa.pub
+              cd /var/www
+              git pull
+              docker pull gcr.io/nlp-flywheel/alchemy:latest && \
+              docker-compose -f docker-compose-staging.yml down && \
```
ShahNewazKhan, line 97: This will cause down time, a way arround this can be to scale up the number of containers to 2 and then remove the oldest container as detailed [here](https://stackoverflow.com/a/54240489/2990066).

cloudbuild.yaml
```diff
@@ -59,6 +59,47 @@ steps:
...
+        gcloud secrets versions access latest --secret=alchemy-builder-ssh > id_rsa
+        gcloud secrets versions access latest --secret=alchemy-builder-ssh-pub > id_rsa.pub
+        echo "" >> id_rsa ; echo "" >> id_rsa.pub ;  # Add a new line to the end
+
+        # INSTANCES=$(gcloud compute instances list --filter="${_DEPLOY_FILTER}")
+        echo Copying the keys
+        gcloud compute ssh \
+            --zone "us-central1-a" \
+            --project "nlp-flywheel" \
+            "${_DEPLOY_MACHINE}" -- "mkdir -p ~/.ssh/"
```
ShahNewazKhan, line 81: Is `_DEPLOY_MACHINE` the VM where the code base is being hosted? I don't understand why the ssh keys are being placed in here and then the repo is being pulled in to a `staging deploy`

MJafarMashhadi: It's the VM in which the website is run and served.
The only reason I'm copying the ssh keys there is to do the `git pull`. And the only reason I'm pulling the whole repository is because I want to have the latest compose config, just for a single file.
It could've been something like `curl https://raw.githubusercontent.com/georgianpartners/annotation_tool/master/docker-compose-staging.yml > docker-compose-staging.yml`, in case we are sure that the repository is going to remain public.

So I am not running the source code directly, but using the latest production image built in cloud build: `gcr.io/nlp-flywheel/alchemy:latest`

cloudbuild.yaml
MJafarMashhadi: It's worse! Not only it causes a downtime for a few seconds/minutes, in case the `docker-compose up` fails it won't roll back to the previous state! I'm doing the `docker pull` before the `compose down` to reduce this downtime. It's not an ideal way of updating the server. However, I talked to @jzhang-gp before implementing it and he said since it is a temporary solution and we will be moving to cloud run and/or GKE very soon we can tolerate these issues.

ShahNewazKhan: Makes sense!

MJafarMashhadi: Btw thanks for the link, I learned something new!

cloudbuild.yaml
ShahNewazKhan: Right, does it make sense to leave the `ssh keys` provisioned on the machine instead of on each pull?

cloudbuild.yaml
ShahNewazKhan: No worries, I'm right there with you!

cloudbuild.yaml
MJafarMashhadi: It makes sense, I can leave them there and delete the key transfer steps here

ShahNewazKhan: Sounds good!

ShahNewazKhan (approved):
LGTM!

jzhang-gp (approved):
LGTM!

PR #178, opened 2020-11-13
Using python client instead of installing gsutils

MJafarMashhadi:


MJafarMashhadi:
The APIs that `gcloud` cli is using are private and prone to have breaking changes, it is discouraged to import and use them. So I think they cannot be replaced at this time. Closing the PR for now.

PR #179, opened 2020-11-16
Update the sorting of the tasks

jzhang-gp:
This is a temporary and urgent request to sort the tasks by name instead of the creation time. Ideally this should be refactored into the TaskDao object but we will have to implement that later. 

Verified on the website. Tasks are now reordered by name. `TestML` and `TestMML` were created way before the other two so the sorting is working.
image


ShahNewazKhan (approved):
LGTM!

MJafarMashhadi (approved):

PR #184, opened 2020-11-26
Update domain names

MJafarMashhadi:
Added `alchemy.impact.georgian.io`


docker-compose-staging.yml
```diff
@@ -74,7 +74,7 @@ services:
     env_file:
       - .env
     environment:
-      - DOMAIN_NAME=alchemy.impact.georgianpartners.com
+      - DOMAIN_NAME=alchemy.impact.georgianpartners.com alchemy.impact.georgian.io
```
jzhang-gp, line 77: Now I think about it, can we move the domain name to a environment variable that's not public? This is an internal domain which is not applicable to everyone else.

MJafarMashhadi: You mean moving it to the `.env` file?

jzhang-gp: If that works without exposing the url in the public, yes.

MJafarMashhadi: I should check to see if it works, nginx container uses the env vars to preprocess the config template to a config file.

MJafarMashhadi: If it worked I'll update this PR to remove this line from compose file

MJafarMashhadi: Worked ^_^ @jzhang-gp

jzhang-gp (approved):
LGTM

PR #185, opened 2020-11-26
Header breadcrumb updated

MJafarMashhadi:
Slightly updated the UI to have a better navigation


jzhang-gp (commented):
Hmm I'm not sure if changing "Home" to "Alchemy" make the navigation better? Are there any confusions you had? 

The change itself is simple. My concern is that our users are used to the Home and Example phrases.

MJafarMashhadi:
My confusion was with the tasks page, it was going from `home/tasks` to `home/task` instead of `home/tasks/task n`. Changing Home to Alchemy was just aesthetic, personal opinion 😅

PR #186, opened 2020-11-30
Let's encrypt SSL

MJafarMashhadi:
I've got the certificates for both staging and production and tested these configs on staging server [https://tube.alchemy.impact.georgian.io/](https://tube.alchemy.impact.georgian.io/).

It's blocking the Okta integration, so I prefer not to wait for @jzhang-gp and I'm asking you for a review. Thanks!

### DIFF
--- a/.conf/nginx/alchemy.impact.georgianpartners.com.conf.template
+++ b/.conf/nginx/alchemy.impact.georgianpartners.com.conf.template
@@ -12,6 +12,7 @@ server {
 
     location ^~ /.well-known/acme-challenge/ {
       default_type "text/plain";
+      autoindex off;
       root /var/www/letsencrypt;
     }
 
--- a/.conf/nginx/alchemy.impact.georgianpartners.com.conf.template
+++ b/.conf/nginx/alchemy.impact.georgianpartners.com.conf.template
@@ -8,8 +8,26 @@ upstream admin_server_us {
 
 server {
     listen 80;
-
     server_name ${DOMAIN_NAME};
+
+    location ^~ /.well-known/acme-challenge/ {
+      default_type "text/plain";
+      root /var/www/letsencrypt;
+    }
+
+    location / {
+        return 301 https://$host$request_uri;
+    }
+}
+
+server {
+    listen 443 ssl;
+    server_name ${DOMAIN_NAME};
+
+    ssl_certificate     /etc/letsencrypt/live/${CERT_NAME}/fullchain.pem;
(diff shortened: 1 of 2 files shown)


.conf/nginx/alchemy.impact.georgianpartners.com.conf.template
```diff
@@ -8,9 +8,27 @@ upstream admin_server_us {
 
 server {
     listen 80;
+    server_name ${DOMAIN_NAME};
+
+    location ^~ /.well-known/acme-challenge/ {
+      default_type "text/plain";
+      root /var/www/letsencrypt;
+    }
```
dhuang-gp, line 16: Is this safe to have available all the time? Shouldn't this be taken down after the cert has been issued? Is there any confidential info in `/var/www/letsencrypt`?

Also, shouldn't the location also include the `<TOKEN>`? Otherwise an attacker could pass any token and get `/var/www/letsencrypt` (assuming there is confidential information there)

MJafarMashhadi: Since it's a `root` directive it'll append the URL, so it'll be serving `/var/www/letsencrypt/.well-known/acme-challenge/` and not the parent folders so it'll be safe. acme-challenge will be empty all the time except for a brief moment when the certificate is being renewed, certbot will put the challenge there, verify the server and then clean the directory up.

Can you explain a bit more about the token issue?

dhuang-gp: Ah, I see. I didn't realize that `/var/www/letsencrypt` is a directory - I thought it was a file. In that case, are directory listings enabled by default on nginx? If yes, I think they should be disabled for this directory.

What I meant by the token is from: https://letsencrypt.org/docs/challenge-types/ where it mentions `http://<YOUR_DOMAIN>/.well-known/acme-challenge/<TOKEN>`. So I was thinking that the web server should only serve requests for `<TOKEN>` and nothing else. But since the file only exists briefly, it's ok.

MJafarMashhadi: Aha, I see! It'll be fine I believe.
Directory listing is off by default in nginx, but I'll add a `autoindex off;` to avoid surprises in case they changed this default in future versions of nginx 😃

dhuang-gp: Thanks!

dhuang-gp (approved):

MJafarMashhadi:
Going to merge after the tests pass, thanks Diego!

PR #187, opened 2020-12-01
Okta integration

MJafarMashhadi:
Must be merged after #186


MJafarMashhadi:
**Migrating all the tasks to the new usernames**

alchemy/shared/okta/saml.py
```diff
@@ -0,0 +1,154 @@
...
+                'authn_requests_signed': False,
+                'logout_requests_signed': True,
+                'want_assertions_signed': True,
+                'want_response_signed': False,
+            },
+        },
+    }
+    spConfig = Saml2Config()
+    spConfig.load(settings)
+    spConfig.allow_unknown_attributes = True
```
jzhang-gp, line 56: I assume the `False`, `allow_unsolicited` and `allow_unknown` configurations are all necessary for the thing to work?

requirements/test.txt
```diff
@@ -1 +1,6 @@
 -r base.txt
+
+# Have both authentication backends while testing
+flask-oidc>=1.4.0
+pysaml2==6.3.1
```
jzhang-gp, line 5: Why do we need both?

tests/okta_mock.py
```diff
@@ -0,0 +1,20 @@
...
+    def update_login_decorator(self, decorator):
+        pass
+
+    def login_required(self, f):
+        @wraps(f)
+        def _wrapper(*args, **kwargs):
+            if self.bypass:
+                return f(*args, **kwargs)
+
+            return redirect('https://dev-8588821.okta.com/oauth2/default/v1/authorize')
```
jzhang-gp, line 19: Anything special about this hardcoded link?

alchemy/admin_server/__init__.py
```diff
@@ -25,22 +24,28 @@
...
-
     if test_config is None:
         # load the instance config, if it exists, when not testing
         app.config.from_object(DevelopmentConfig)
     else:
         # load the test config if passed in
         app.config.from_object(test_config)
 
+    app.config.update({
+        'SECRET_KEY': 'athena_todo_change_this_in_prod',
```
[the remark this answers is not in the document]
MJafarMashhadi: It's used for encrypting session cookie, it should be read from a secret manager or env variables

alchemy/admin_server/templates/tasks/show.html
```diff
@@ -170,20 +170,6 @@ <h4>
...
-            <ul>
-              {% for user, link in annotator_login_links %}
-              <li>
-                <b>{{user}}</b>: <span onclick="selectText(this.id)" id="{{link}}">{{link}}</span>
-              </li>
-              {% endfor %}
-            </ul>
-          </div>
-        </div>
-
```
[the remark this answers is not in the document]
MJafarMashhadi: So all the links will be something like `alchemy.io/tasks/5`?

requirements/test.txt
MJafarMashhadi: I didn't remove the OpenID code, so other people who are not using SAML can still use that. It's only in the testing image and will not be installed in production or staging images.

tests/okta_mock.py
MJafarMashhadi: Hmmm could be `anything.okta.com/oauth2/....`

alchemy/admin_server/__init__.py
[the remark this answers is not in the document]
MJafarMashhadi: f398301

alchemy/shared/okta/saml.py
MJafarMashhadi: I need to double check that; I copied the parameters from their example code

alchemy/db/model.py
```diff
@@ -128,7 +129,7 @@ class AnnotationValue:
 # Tables
 
 
-class User(Base):
+class User(Base, flask_login.UserMixin):
```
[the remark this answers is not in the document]
MJafarMashhadi: For flask_login package. It abstracts the user as a class; i.e. when I'm calling the `login` or `logout` methods, it expects the parameter that's passed in to implement `UserMixin`'s interface. I think it is a good idea to use the same user model that we already have instead of having two separate classes for managing user accounts

alchemy/annotation_server/tasks.py
```diff
@@ -29,15 +29,15 @@
...
 
 bp = Blueprint("tasks", __name__, url_prefix="/tasks")
 
 
 @bp.route("/<string:id>")
-@login_required
+@auth.login_required
 def show(id):
-    username = g.user["username"]
+    username = g.user['email']
```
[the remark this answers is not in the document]
MJafarMashhadi: g.user is a dictionary with very arbitrary structure, I think it's better to refactor it to be the actual `User` class, for the same reason as this comment: https://github.com/georgianpartners/annotation_tool/pull/187#discussion_r541026592

I had to make this change after openID connect but didn't roll it back when switched to SAML

alchemy/annotation_server/templates/index.html
```diff
@@ -2,7 +2,7 @@
 
 {% block content %}
 <h2>
-  Hi {% if g.user %} {{g.user['username']}} {% endif %}
+  Hi {% if g.user %} {{ g.user['name'] }} {% endif %}
```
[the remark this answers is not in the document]
MJafarMashhadi: With openid connect we can access to the actual first name and last name of the users

alchemy/admin_server/templates/tasks/show.html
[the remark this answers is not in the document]
MJafarMashhadi: Addressed in 9b9b473

alembic/versions/22da894fc7ba_data_migration.py
```diff
@@ -0,0 +1,170 @@
+"""data_migration
```
jzhang-gp, line 1: Should we move this script to the script folder? Even though it is doing data migration, it's not the same as the rest of the alembic script which handles database schema changes. 

This is more of a one-time data change and is not going to be executed every time we do data migration through `alembic upgrade`. 

Besides, can we move the mappings of people's username to a separate config file on your local machine and exclude the config file from the PR? 

The usernames are our own data (even if they are public) that should not be included in a public repo.

MJafarMashhadi: It makes sense, I can rebase the commits to remove them from the git history

alchemy/shared/okta/saml.py
MJafarMashhadi: `allow_unsolicited` allows the users to click on Alchemy's link in Okta's dashboard to authenticate and log into alchemy; i.e. the authentication can start in ways other than clicking login button in Alchemy's webpage. So It's not a necessary part, but can be nice to have.

MJafarMashhadi: https://github.com/IdentityPython/pysaml2/blob/master/docs/howto/config.rst#authn-requests-signed

MJafarMashhadi: Allow unkown attributes was necessary at some point but I commented it out now and it's still working, I'll update the code

MJafarMashhadi: Fixed the allow_unknown in a8042443d

The commit looks small but just figuring those parameters out drove me crazy 😅😂

.github/workflows/ci.yaml
```diff
@@ -28,6 +28,7 @@ jobs:
 
         echo "ALCHEMY_DATABASE_URI=\"sqlite://$RUNNER_TEMP/alchemy.db\"" >> $GITHUB_ENV
         echo "ALCHEMY_FILESTORE_DIR=\"$RUNNER_TEMP/__filestore\"" >> $GITHUB_ENV
+        echo "SECRET_KEY=randomsecretkeytopreventerrors" >> $GITHUB_ENV
```
jzhang-gp, line 31: Does this expose any security issues?

MJafarMashhadi: It's just a placeholder, no risk 😅

jzhang-gp (approved):
LGTM

PR #190, opened 2020-12-15
Created migration script

MJafarMashhadi:
Pulled the changes out of `okta_integration` branch


jzhang-gp (approved):
As discussed offline, the script has been tested locally. We should do a merge test on the staging database again during deployment and migration.

PR #192, opened 2020-12-15
Open up postgres port on local

MJafarMashhadi:
Open up database port on local development env (docker-compose)


jzhang-gp (approved):
LGTM as long as this is only used in local.

MJafarMashhadi:
This file is only used in local development. In staging (and production for now) docker-compose-staging.yml is used, which doesn't have the DB service; they connect to the respective DB instances on the cloud.

PR #193, opened 2020-12-15
Raise 404 for invalid IDs in create/edit task page

MJafarMashhadi:
Raise 404 for invalid IDs in create/edit task page (currently it'll result in null pointer exceptions and a 500 response)


jzhang-gp (approved):
LGTM. Are the changes aside from the 404 code just formatting changes?

MJafarMashhadi:
> LGTM. Are the changes aside from the 404 code just formatting changes?

It was an if statemenet similar to:

```python
if condition:
    # ...tens of lines of code...
    return A

return error```

I changed it to:

```python
if not condition:
    return error

# ...tens of lines of code...
return A
```

The change looks large because the change in indentation.

PR #195, opened 2020-12-15
Update task annotators input to only accept currently registered users

MJafarMashhadi:
Before:
Screen Shot 2020-12-15 at 4 30 53 PM

After:
Screen Shot 2020-12-15 at 4 31 39 PM


alchemy/admin_server/static/select2-bootstrap4.min.css
```diff
@@ -0,0 +1 @@
...-dropdown.select2-dropdown--above{border-top:1px solid #ced4da;border-top-left-radius:.25rem;border-top-right-radius:.25rem}.select2-container--bootstrap4 .select2-dropdown .select2-results__option[aria-selected=true]{background-color:#e9ecef}.select2-container--bootstrap4 .select2-results__option--highlighted,.select2-container--bootstrap4 .select2-results__option--highlighted.select2-results__option[aria-selected=true]{background-color:#007bff;color:#f8f9fa}.select2-container--bootstrap4 .select2-results__option[role=group]{padding:0}.select2-container--bootstrap4 .select2-results__option[role=group] .select2-results__options--nested .select2-results__option{padding-left:1em}.select2-container--bootstrap4 .select2-results>.select2-results__options{max-height:15em;overflow-y:auto}.select2-container--bootstrap4 .select2-results__group{padding:6px;display:list-item;color:#6c757d}.select2-container--bootstrap4 .select2-selection__clear{width:.9em;height:.9em;line-height:.75em;padding-left:.15em;margin-top:.7em;border-radius:100%;background-color:#c8c8c8;color:#f8f9fa;float:right;margin-right:.3em}.select2-container--bootstrap4 .select2-selection__clear:hover{background-color:#afafaf}
```
jzhang-gp, line 1: I assume this file is necessary for the UI change? Where did we get this file from?

alchemy/admin_server/tasks.py
```diff
@@ -478,17 +503,13 @@ def parse_labels(form):
 
 
 def parse_annotators(form):
-    annotators = form["annotators"]
-    assert annotators, "Annotators is required"
+    annotators = [int(uid) for uid in form.getlist("annotators[]")]
+    assert annotators and len(annotators) > 0, "You should select at least one person to annotate"  # but why?
```
jzhang-gp, line 507: "but why?" What concerns do you have? Does it not make sense to have at least one annotator for the task?

alchemy/admin_server/templates/tasks/edit.html
```diff
@@ -1,5 +1,13 @@
 {% extends 'base.html' %}
 
+{% block head %}
+  <link href="https://cdnjs.cloudflare.com/ajax/libs/select2/4.0.6-rc.0/css/select2.min.css" rel="stylesheet" />
+  <link href="/static/select2-bootstrap4.min.css" rel="stylesheet" />
+
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/select2/4.0.6-rc.0/js/select2.min.js"></script>
```
jzhang-gp, line 7: Is this the recommended way to include js for the select menu? My concern is that we are depending on cloudflare...

alchemy/admin_server/templates/tasks/new.html
```diff
@@ -1,5 +1,13 @@
 {% extends 'base.html' %}
 
+{% block head %}
+  <link href="https://cdnjs.cloudflare.com/ajax/libs/select2/4.0.6-rc.0/css/select2.min.css" rel="stylesheet" />
+  <link href="/static/select2-bootstrap4.min.css" rel="stylesheet" />
+
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/select2/4.0.6-rc.0/js/select2.min.js"></script>
```
jzhang-gp, line 7: Same here.

alchemy/admin_server/tasks.py
```diff
@@ -80,16 +77,23 @@ def index():
...
+        users=users,
     )
 
 
 @bp.route("/", methods=["POST"])
 def create():
     data_fnames = get_all_data_files()
+    users = db.session.query(User).all()
+    for u in users:
+        u.is_selected = False
```
jzhang-gp, line 96: `is_selected` is a dynamic property added on the fly, right? Is there another way to limit the set of the users?

alchemy/admin_server/static/select2-bootstrap4.min.css
MJafarMashhadi: Yes, This CSS file is from https://github.com/ttskch/select2-bootstrap4-theme that makes select2 work with bootstrap4. Select2 originally only works with bootstrap up to version 3.

alchemy/admin_server/templates/tasks/edit.html
MJafarMashhadi: What about it is concerning?

alchemy/admin_server/tasks.py
MJafarMashhadi: Not a concern. Just in case that a new person should make annotations but they don't have a user account (yet) for some reason, why shouldn't we be able to create a task, like a draft mode.
It's too hypothetical now that I think about it 😅

alchemy/admin_server/tasks.py
MJafarMashhadi: Yes it's added to only be used in the template. Won't change things in other places.

I think it might be possible to pass the set of 'selected' users to the template and change the if expression from `{% if user.is_selected %}blah` to `{% if user.username in selected_annotators %} blah`. I'll try that, it makes more sense.

MJafarMashhadi: Done

alchemy/admin_server/templates/tasks/new.html
MJafarMashhadi: #TechDebt haha

alchemy/admin_server/tasks.py
jzhang-gp: Ah ok. I don't think we need to worry about that use case at this moment. Can you remove the comment?

alchemy/admin_server/templates/tasks/edit.html
jzhang-gp: My initial concern was that this might not be the official way to integrate the JS but after checking their [github repo](https://github.com/select2/select2), it seems that this is the recommended way.

jzhang-gp (approved):
LGTM

alchemy/admin_server/tasks.py
MJafarMashhadi: Sure

PR #196, opened 2020-12-16
Annotation server warn expired login link

MJafarMashhadi:
Related to #195


MJafarMashhadi:
No longer required. It was a intermediate step before bringing Okta into production.

PR #198, opened 2021-01-06
AUPR changes in the training script

jzhang-gp:
Change the metric calculation to AUPR based on the discussion on Slack


alchemy/train/no_deps/metrics.py
```diff
@@ -2,7 +2,7 @@
 from sklearn.metrics import (
     confusion_matrix,
     precision_recall_fscore_support,
-    roc_auc_score,
+    average_precision_score,
```
MJafarMashhadi, line 5: Maybe we can have a dynamic way to add/remove/update the metrics as a future feature. sklearn has already done much of the heavy lifting on our behalf, i.e. creating a unified interface for all the metrics, we just need to do some basic `importlib` magic to load the metrics dynamically and apply them on the data.

MJafarMashhadi (approved):

angeliney:
@jzhang-gp can we add the AUPR and still keep the previous ROC AUC?

angeliney (changes requested):
As mentioned in an earlier comment, let's keep the original roc_auc_score but add the average_precision_score on top :) Is that possible?

jzhang-gp:
> @jzhang-gp can we add the AUPR and still keep the previous ROC AUC?

You mean add it as a separate metric? 

> @jzhang-gp can we add the AUPR and still keep the previous ROC AUC?



> As mentioned in an earlier comment, let's keep the original roc_auc_score but add the average_precision_score on top :) Is that possible?

Sure. We can do that.

jzhang-gp:
image
@angeliney does this look OK to you?

angeliney (approved):
Looks good!!

PR #199, opened 2021-01-07
Address the comments in review

MJafarMashhadi:


jzhang-gp (commented):
Have you tested triggering the training process?

MJafarMashhadi:
> Have you tested triggering the training process?

I haven't, do you have an example file handy that I can use? If you don't I will download some data myself and test it

MJafarMashhadi:
> > Have you tested triggering the training process?
> 
> I haven't, do you have an example file handy that I can use? If you don't I will download some data myself and test it

Still works, nothing is broken

jzhang-gp (approved):
LGTM

PR #201, opened 2021-01-08
Show users' names in tasks page instead of the raw usernames

MJafarMashhadi:
These 6 places are updated:
image

image

4 places here:
image


jzhang-gp (approved):
LGTM

PR #202, opened 2021-01-08
Use select2 in bulk upload page

MJafarMashhadi:
Updated bulk upload pages to restrict user names to the ones already defined.

https://georgian.slack.com/archives/G0171SNNALX/p1610131772035900
https://app.clickup.com/t/1236149/SA-1871

### DIFF
--- a/alchemy/admin_server/annotations.py
+++ b/alchemy/admin_server/annotations.py
@@ -47,6 +47,7 @@ def bulk_upload_positive_annotations():
         "annotations/bulk_upload_positive_annotations.html",
         annotation_values=acceptable_values,
         redirect_to=request.referrer,
+        users=db.session.query(User).all(),
         entity_types=EntityTypeEnum.get_all_entity_types(),
     )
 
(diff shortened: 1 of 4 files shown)


alchemy/admin_server/annotations.py
```diff
@@ -47,6 +47,7 @@ def bulk_upload_positive_annotations():
         "annotations/bulk_upload_positive_annotations.html",
         annotation_values=acceptable_values,
         redirect_to=request.referrer,
+        users=db.session.query(User).all(),
```
jzhang-gp, line 50: If they are changes within the same function, can we create one variable to store the users instead of repeating the queries?

jzhang-gp (approved):
LGTM

alchemy/admin_server/annotations.py
MJafarMashhadi: I'll double check, but doesn't SQLAlchemy cache that query?

MJafarMashhadi: They're in separate functions

PR #189, opened 2020-12-01
Redirect to georgian.io

MJafarMashhadi:
`https?://*.georgianparners.com/*` will be mapped to `https://$1.georgian.io/$2`

PR #204, opened 2021-01-11
Put the config line back in the nginx service in docker-compose

MJafarMashhadi:
This line was deleted by mistake in #189 reverting that change.


jzhang-gp (approved):
LGTM.

PR #205, opened 2021-01-12
SA 1581 Setting up alarms or alerts to track errors

MJafarMashhadi:
Added non-authenticated --and cached-- endpoints for a basic status check endpoint.

### DIFF
--- a/alchemy/shared/health_check.py
+++ b/alchemy/shared/health_check.py
@@ -29,7 +29,10 @@ def check_file_system() -> bool:
 def check_celery() -> bool:
     try:
         from alchemy.ar.ar_celery import hello
-        task_result = hello.apply_async()
+        task_result = hello.apply_async(retry=False, retry_policy={
+            'timeout': 5.0,
+            'max_retries': 1,
+        })
         task_result.wait(timeout=5)
         assert task_result.successful()
     except:
--- a/alchemy/shared/health_check.py
+++ b/alchemy/shared/health_check.py
@@ -1,0 +1,38 @@
+from datetime import datetime
+
+from alchemy.db.fs import models_dir
+from alchemy.db.utils import get_all_data_files
+
+
+def check_file_system() -> bool:
+    # Check read
+    try:
+        get_all_data_files()
+    except:
+        return False
+
+    # Check write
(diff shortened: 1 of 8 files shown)


alchemy/shared/health_check.py
```diff
@@ -0,0 +1,33 @@
+from datetime import datetime
+
+from alchemy.db.fs import models_dir
+from alchemy.db.utils import get_all_data_files
+
+
+def check_file_system():
```
jzhang-gp, line 7: Is it part of the best practice to check file system in health ping check?

alchemy/shared/health_check.py
```diff
@@ -0,0 +1,33 @@
...
+        data_files = get_all_data_files()
+    except:
+        return 'error'
+
+    # Check write
+    try:
+        f = (models_dir(as_path=True) / datetime.utcnow().strftime("test-%Y%m%d%H%M%S.txt"))
+        f.write_text("test", encoding='utf-8', errors='ignore')
+    except:
+        return 'error'
```
jzhang-gp, line 19: Would this creates a small file every time we do a check? That could overflow the disk at some point, right?

MJafarMashhadi: That's right, I wanted to delete the file in the end, I put the variable `f` there for that (otherwise I would just put all of them in one line). For some reason I forgot to do it. Good thing that we have these reviews lol

alchemy/shared/health_check.py
MJafarMashhadi: I haven't explicitly seen any recommendations to check if file system is read/writable, but since it has happened once that the fs was mounted read-only I put it in health check as well.

jzhang-gp: Hmm in this case would it be excessive to check the read/write of the disk on every ping? A one-time check after deployment should be enough right? Read/write may be something worth checking but probably not here. Can we remove the read/write check here and move it somewhere else?

MJafarMashhadi: What about runserver script?

MJafarMashhadi: Moved it to the runserver script `1d9fa7d`

PR #209, opened 2021-01-16
Serving static files directly from nginx

MJafarMashhadi:


docker-compose-staging.yml
```diff
@@ -1,22 +1,21 @@
...
-
   admin_server:
     env_file: .env
     image: &img gcr.io/nlp-flywheel/alchemy:latest
     expose:
       - 5000
-    volumes: *default-volumes
+    volumes:
+      - $GCLOUD_LOCAL_DIR:/root/.config/gcloud
+      - $FILESTORE_LOCAL_DIR:/app/__filestore
```
jzhang-gp, line 15: We are duplicating these lines in multiple places. Is this the only way?

MJafarMashhadi: It used to be defined once on top of the file and referenced down in the services, using a & * syntax like C pointers, but that only works when you have the exact same array. You can't reference an array and then append a few items to it; Apparently yaml [doesn't (and won't)](https://github.com/yaml/yaml/issues/35) support merging arrays.

jzhang-gp (approved):
LGTM

MJafarMashhadi:
Thanks!

PR #212, opened 2021-01-25
SA-3034 Added basic login backend for local testing

MJafarMashhadi:


alchemy/shared/okta/basic_auth.py
```diff
@@ -0,0 +1,62 @@
...
+    return None
+
+
+def _create_blueprint():
+    bp = flask.Blueprint('auth', 'basic_auth')
+
+    # Start log in
+    @bp.route('/login', methods=['GET', 'POST'])
+    def login():
+        error = False
```
MJafarMashhadi, line 26: Should remove this variable.

jzhang-gp: Which variable and is it removed?

alchemy/shared/okta/basic_auth.py
```diff
@@ -0,0 +1,62 @@
+import flask
```
jzhang-gp, line 1: So how do we ensure the basic login is used during local testing? Do we need any configuration for that?

Besides, do these file depend on anything else in the `okta` folder? Should we move it to a separate path?

alchemy/shared/okta/basic_auth.py
MJafarMashhadi: `error`, not yet

alchemy/shared/okta/basic_auth.py
MJafarMashhadi: Manually changing the respective config value in local (since we don't have separate config files for local, production, etc)

MJafarMashhadi: The folder needs to be renamed to something else such as `auth_backends` since it's not okta-related-only anymore

alchemy/shared/okta/basic_auth.py
MJafarMashhadi: now done

alchemy/shared/okta/basic_auth.py
MJafarMashhadi: updated

jzhang-gp: > Manually changing the respective config value in local (since we don't have separate config files for local, production, etc)

Can you elaborate on which config we need to change?

MJafarMashhadi: in `admin_server/__init__.py` and `annotation_server/__init__.py` this line needs to change:

```python
'AUTH_BACKEND': 'alchemy.shared.auth_backends.saml',
```
to
```python
'AUTH_BACKEND': 'alchemy.shared.auth_backends.basic_auth',
```

locally

jzhang-gp: Cool, I'll approve the change. Can you add this as a comment in the code or in the README?

jzhang-gp (approved):
LGTM. As mentioned, please add a comment in the README to explain how to change login method in the local env.

alchemy/shared/okta/basic_auth.py
MJafarMashhadi: Will add, thanks!

balmasi:
Task linked: [SA-3034 As an Engineer I want a hassle-free way of testing user authentication on my local dev environment](https://app.clickup.com/t/1236149/SA-3034)

PR #215, opened 2021-02-03
SA-3149 Test and fix other problems that caused the infinite redirect loop

MJafarMashhadi:


alchemy/shared/auth_backends/saml.py
```diff
@@ -182,21 +183,22 @@ def login():
...
-        annotation_server_url = Config.get_annotation_server()
-        if _drop_protocol(annotation_server_url) == _drop_protocol(flask.request.url_root):
+        def _strip_url(url):
+            # Removes the protocol and the trailing slash.
+            # Assumes there is no query string
+            return re.sub("^([^:]+://)?(([^/]+/)*)([^/]+)(/?)", r"\2\4", url.strip())
+
+        annotation_server_url = _strip_url(Config.get_annotation_server())
+        url_root = _strip_url(flask.request.url_root)
+        if annotation_server_url == url_root:
```
jzhang-gp, line 193: Hmm do you think we can leverage the package [`tldextract`](https://github.com/john-kurkowski/tldextract) instead of using a regular expression?

MJafarMashhadi: I was so rushed to just get this over with that I forgot about this: https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse

jzhang-gp (approved):
LGTM

PR #217, opened 2021-02-16
SA-3066 Change the dependency installation process and fix some typos

MJafarMashhadi:
* Cleaning the apt cache is now postponed to creating the production images, it has two benefits:
  - It's easier to install new packages for debugging in the local development image (such as vim, ps, pkill, etc)
  - Removed the need for an extra `apt update` when installing production dependencies, making the build ~15-20 seconds faster.
* Removed the command that deletes *.pyc files. They are going to be recreated on the first run, so it makes the cold start a bit faster
* Two typos are fixed

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,8 +5,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
          wget \
          curl \
          git \
-         xmlsec1 && \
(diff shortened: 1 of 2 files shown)


jzhang-gp (approved):
LGTM

PR #218, opened 2021-02-17
Quick fix of the model id change issue

jzhang-gp:
This change is to resolve the issue that when the model id has changed, users may see models with duplicate version numbers. In fact, they are models under different model ids, even though the version id is the same. This has caused some confusion. 

The quick fix is to organize the model information under different model id then version number. 

Since the model id is a bit long, for the sake of the UI, we shortened it to 7 characters, like the shortened git commit number. 7 characters have enough combinations (36^7) for our use case.

image

image

### DIFF
--- a/alchemy/admin_server/models.py
+++ b/alchemy/admin_server/models.py
@@ -233,12 +233,8 @@ def _collect_model_data_rows():
 
         row = ModelDataRow(
             label=label,
-            latest_version=latest_model.uuid[:7] + "-V" + str(latest_model.version)
-            if latest_model
-            else None,
-            deployed_version=deployed_model.uuid[:7]
-            + "-V"
-            + str(deployed_model.version)
+            latest_version=latest_model.full_version_name if latest_model else None,
+            deployed_version=deployed_model.full_version_name
             if deployed_model
             else None,
             num_of_data_points=chosen_model.get_len_data() if chosen_model else None,
--- a/alchemy/admin_server/models.py
+++ b/alchemy/admin_server/models.py
@@ -26,8 +26,8 @@ bp = Blueprint("models", __name__, url_prefix="/models")
 @dataclass
 class ModelDataRow:
     label: str
-    latest_version: int
-    deployed_version: int
+    latest_version: str
+    deployed_version: str
     num_of_data_points: int
     threshold: float
     majority_annotator: str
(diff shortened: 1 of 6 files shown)


alchemy/admin_server/templates/tasks/show.html
```diff
@@ -365,258 +365,268 @@ <h4>
...
+                        <h5>Model ID: {{ uuid[:7] }}</h5>
+                    </div>
+                    {% for mv in models_by_uuid.get(uuid, []) %}
+
+                    <div class="card mt-4">
+                      <div class="card-body">
+                        <h5>
+                          Version {{mv.version}}
+                          ({{mv.created_at.strftime('%Y-%m-%d')}})
+                          <small class='text-muted ml-3'>{{mv.uuid[:7]}}</small>
```
MJafarMashhadi, line 382: `{{ mv.short_uuid }}`

alchemy/admin_server/templates/models/index.html
```diff
@@ -29,8 +29,8 @@ <h4>
                     <tbody>
                       <tr>
                           <th>Label</th>
-                          <th>Latest Version</th>
-                          <th>Deployed Version</th>
+                          <th>Latest Model</th>
+                          <th>Deployed Model</th>
```
MJafarMashhadi, line 33: This is actually for #L48-49, I suggest using `default` values to make UI more friendly:
```html
                        <td>{{ row.latest_version | default('-') }}</td>
                        <td>{{ row.deployed_version | default('Not deployed') }}</td>
```

MJafarMashhadi (approved):
Comments are just suggestions, you might decide to not apply them, so 'Approve' rather than  'Request changes'.

alchemy/admin_server/models.py
```diff
@@ -233,8 +233,14 @@ def _collect_model_data_rows():
 
         row = ModelDataRow(
             label=label,
-            latest_version=latest_model.version if latest_model else None,
-            deployed_version=deployed_model.version if deployed_model else None,
+            latest_version=latest_model.uuid[:7] + "-V" + str(latest_model.version)
```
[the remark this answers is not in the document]
jzhang-gp: good point :)

alchemy/admin_server/templates/tasks/show.html
jzhang-gp: Done.

alchemy/admin_server/templates/models/index.html
jzhang-gp: This one I'll keep it for now since it requires a few other small changes to make things consistent.

PR #220, opened 2021-02-18
SA-3231 Change the if condition + added comment

MJafarMashhadi:
Instead of unsetting `_NO_DEPLOY` on cloud build, one can now just set it to 0 to enable automatic deployments. 

#172 related.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -68,7 +68,8 @@ steps:
   args:
     - "-c"
     - |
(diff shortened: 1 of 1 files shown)


jzhang-gp (approved):
LGTM

balmasi:
Task linked: [SA-3231 As an Engineer I want to enable automatic deployment for Alchemy by setting _NO_DEPLOY to 0 instead of deleting it](https://app.clickup.com/t/1236149/SA-3231)

PR #221, opened 2021-02-26
SA-3317 Updated nginx configurations to facilitate SSL renewal with no downtime.

MJafarMashhadi:
The configuration was missing required parts for renewing the old domain certificates. It worked (and works) fine for the new domain after rebranding.


jzhang-gp (approved):
LGTM

PR #223, opened 2021-03-04
bytes -> str

MJafarMashhadi:
Improved logs


jzhang-gp (commented):
Should we use Logger instead of print? Or is it required by GCP to use print?

MJafarMashhadi:
@jzhang-gp There's no difference for now, they'll both end up in stdout which is piped to google cloud logging (in docker-compose only). However, one thing I had in mind to do when migrating to Django was to replace all `print` statements with appropriate loggers.

PR #224, opened 2021-03-08
Limit celery autoscale

MJafarMashhadi:
* Change upper limit of autoscale from 100 to 10, since 100 is too much for a single machine deployment
* Added restart policies for the web servers and Nginx


jzhang-gp (approved):
LGTM

